PR #1, opened 2013-03-02
Allow building resiprocate from an external directory

joegen:
...ource top dir.   This patch allows building from an external directory so that object files and other files generated by automake and autoconf do not pollute the source tree.

Example:
git clone https://github.com/resiprocate/resiprocate
cd resiprocate
autoreconf -if
cd ..
mkdir BUILD
cd BUILD
../resiprocate/configure --prefix=`pwd`/build
make && make install

### DIFF
--- a/b2bua/Makefile.am
+++ b/b2bua/Makefile.am
@@ -5,6 +5,7 @@ EXTRA_DIST = README.txt
 SUBDIRS = . 
 
 #AM_CXXFLAGS = -DUSE_ARES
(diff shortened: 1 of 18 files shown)


dpocock:
This was attempted with commit 728e297acb8ff826a903a4e3094bc778096b55c5

It looks like it could be improved though:

- to make it work for both C and C++, should be using AM_CPPFLAGS instead of AM_CXXFLAGS
- may need to use $(top_builddir) in *_LDADD

PR #10, opened 2014-10-28
Resiprocate 1.9

qianqiuwanzi:
Some projects builds failed on VS2013.
Why not update sln to VS2013?


dpocock:
It appears you have opened a pull request when you just wanted to ask a question.  Please use the mailing list for questions like this http://list.resiprocate.org/mailman/listinfo/resiprocate-users

PR #12, opened 2015-01-16
Select HashMap implememnation based on libstdc++ version

gjasny:
The current implementation performs the selction based on the
compiler version. When mixing GCC and Clang compiled objects
on RHEL6 this leads to ABI incompatible objects because Clang
pretends to be GCC 4.2 and the system GCC is 4.4.

This patch performs the decision based on the GNU libstdc++
version which is being used. This ensures the same HashMap
implementation across different compilers on a system which
share the same stdlib.

Signed-off-by: Gregor Jasny <Gregor.Jasny@citrix.com>
Reviewed-by: Jupp Mueller <Jupp.Mueller@citrix.com>

### DIFF
--- a/rutil/HashMap.hxx
+++ b/rutil/HashMap.hxx
@@ -11,7 +11,15 @@
    @ingroup data_structures
 */
 
-#  if ( (__GNUC__ == 4) && (__GNUC_MINOR__ >= 3) ) || ( __GNUC__ > 4 )
+// For GNU libstdc++ the decision which hashmap implementation to use must be
+// done on the version of libstdc++ and not on the compiler version.
+// Otherwise for example mixing RHEL6 GCC (4.4.7) and Clang 3.4 on RHEL6 will produce
+// ABI incompatible results.
+// The version encodings are described here:
+//   https://gcc.gnu.org/onlinedocs/libstdc++/manual/abi.html
+#include <map> // force include of nonstandard <bits/c++config.h>
+
+#  if ( defined(__GLIBCXX__) && (__GLIBCXX__ >= 20080306) ) // >= 4.3.0
 #    include <tr1/unordered_map>
 #    include <tr1/unordered_set>
 #    define HASH_MAP_NAMESPACE std::tr1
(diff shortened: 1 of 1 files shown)


pww71:
Work to do a good job, must first sharpen his tools.
Here are some examples that can make you understand.
Hi
My English is not good and could you help me check this document error
Welcome to my space:
http://sourceforge.net/p/pwwhashmap/discussion/general/

Title:       The core of the core of the big data solutions -- Map
Author:      pengwenwei
Email:       wenwei19710430@yahoo.com 
Language:    c++
Platform:    Windows, linux
Technology:  Perfect hash algorithm
Level:       Advanced
Description: Map algorithm with high performance
Section      MFC c++ map stl
SubSection   c++ algorithm
License:     (GPLv3)

    Download demo project - 1070 Kb
    Download source - 1070 Kb

Introduction:
For the c++ program, map is used everywhere.And bottleneck of program performance is often the performance of map.Especially in the case of large data,and the business association closely and unable to realize the data distribution and parallel processing condition.So the performance of map becomes the key technology.

In the work experience with telecommunications industry and the information security industry, I was dealing with the big bottom data,especially the most complex information security industry data,all can’t do without map.

For example, IP table, MAC table, telephone number list, domain name resolution table, ID number table query, the Trojan horse virus characteristic code of cloud killing etc..

The map of STL library using binary chop, its has the worst performance.Google Hash map has the optimal performance and memory at present, but it has repeated collision probability.Now the big data rarely use a collision probability map,especially relating to fees, can’t be wrong.

Now I put my algorithms out here,there are three kinds of map,after the build is Hash map.We can test the comparison,my algorithm has the zero probability of collision,but its performance is also better than the hash algorithm, even its ordinary performance has no much difference with Google.

My algorithm is perfect hash algorithm,its key index and the principle of compression algorithm is out of the ordinary,the most important is a completely different structure,so the key index compression  is fundamentally different.The most direct benefit for program is that for the original map need ten servers for solutions but now I only need one server.
Declare: the code can not be used for commercial purposes, if for commercial applications,you can contact me with QQ 75293192.
Download:
https://sourceforge.net/projects/pwwhashmap/files

Applications:
First,modern warfare can’t be without the mass of information query, if the query of enemy target information slows down a second, it could lead to the delaying fighter, leading to failure of the entire war. Information retrieval is inseparable from the map, if military products use pwwhashMap instead of the traditional map,you must be the winner.

Second,the performance of the router determines the surfing speed, just replace open source router code map for pwwHashMap, its speed can increase ten times.
There are many tables to query and set in the router DHCP ptotocol,such as IP,Mac ,and all these are completed by map.But until now,all map are  using STL liabrary,its performance is very low,and using the Hash map has error probability,so it can only use multi router packet dispersion treatment.If using pwwHashMap, you can save at least ten sets of equipment.

Third,Hadoop is recognized as the big data solutions at present,and its most fundamental thing is super heavy use of the map,instead of SQL and table.Hadoop assumes the huge amounts of data so that the data is completely unable to move, people must carry on the data analysis in the local.But as long as the open source Hadoop code of the map changes into pwwHashMap, the performance will increase hundredfold without any problems.


Background to this article that may be useful such as an introduction to the basic ideas presented:
http://blog.csdn.net/chixinmuzi/article/details/1727195

dpocock:
Thanks for the contribution.

PR #13, opened 2015-04-21
Merge latest work from Mozilla.

docfaraday:


dpocock:
Hi Byron, I think you can just go ahead and merge this directly, you have access to the repository.  If there is some specific part of the change that you would like somebody to review, please let us know.  v1.10.0 is coming soon, so please think about whether this should be merged before or after the release branch is created.

docfaraday:
I see no harm in merging this now.

PR #14, opened 2015-04-23
Update the Wants and After details for the repro service

TafThorne:
Service fails to begin on Fedora 21, this seems to work around the problem.  I started a conversation with Daniel Pocock via the repro-users distribution list and he asked that I provide a pull request so that people can discuss the pros and cons of this particular solution in the pull request.

### DIFF
--- a/repro/pkg/fedora/repro.service
+++ b/repro/pkg/fedora/repro.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=reSIProcate repro SIP proxy daemon
+After=network-online.target
 After=syslog.target
-After=network.target
+Wants=network-online.target
 
 [Service]
 ExecStart=/usr/sbin/repro /etc/repro/repro.config --Daemonize=false


dpocock:
I just had a look at:

http://www.freedesktop.org/wiki/Software/systemd/NetworkTarget/

It appears that another option exists for people who have this problem, executing the command

    systemctl enable systemd-networkd-wait-online.service

should ensure that whenever you boot, the system waits for IP addresses to be ready before starting other things.  This is a workaround that may be valid for users of the existing packages.

The actual demand for this patch varies.  For people binding to a fixed IP, it is quite important that the IP should be available before the service starts.  For people binding to 0.0.0.0, it is not important to wait for the DHCP.  Given that SIP works better with specified IP addresses, I am in favour of this patch and so I'm going to accept it.

People who don't want to wait for new versions of the package can use the workaround above.

dpocock:
Discussed here: http://list.resiprocate.org/archive/repro-users/msg00910.html

PR #15, opened 2015-04-23
Update the Wants and After details for the reTurnServer service

TafThorne:
I noticed that the repro service was not starting up as it should do n Fedora 21 for me.  I discussed this via the repro-users mailing list with Daniel Pocock and he asked that I generate a patch and pull request for both the repro and reTurnServer so that the pros and cons of this particular solution could be discussed in the pull request.  

These changes made the repro.service start reliably on reboot for me.  Presumably the same changes would work for the reTurn service too.

### DIFF
--- a/reTurn/pkg/fedora/resiprocate-turn-server.service
+++ b/reTurn/pkg/fedora/resiprocate-turn-server.service
@@ -1,7 +1,8 @@
 [Unit]
 Description=reSIProcate reTurn ICE/STUN/TURN server daemon
+After=network-online.target
 After=syslog.target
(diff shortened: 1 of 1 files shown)


dpocock:
Discussed here: http://list.resiprocate.org/archive/repro-users/msg00910.html

Merging this patch based on the same reasons given in #14 

The reasons for merging this are even more compelling for STUN/TURN than for SIP, as a STUN/TURN server should never really be running on 0.0.0.0.

Thomas, thanks for contributing this feedback and the pull requests.

PR #25, opened 2015-07-22
Innocent change of a helper

apronchenkov:


TafThorne:
> Innocent change of a helper.

Well that is a worrying commit comment if I ever saw one.  Looks like it should not have compiled before.  If this is obfuscated evil code it is a good one.

dbozzali:
In ServerAuthManager::handleUserAuthInfo() method it is used static Helper::advancedAuthenticateRequest() that doesn't have the issue since it uses Helper::makeResponseMD5WithA1() with the proper number of arguments.
In this way the change is innocent (and maybe the fixed mistake).
You could face the issue if you use Helper::authenticateRequest() to check authentication and you want to use qop, since in that case (before the patch) you use Helper::makeResponseMD5() actually without taking into account message contents...
@TafThorne Nice comment :)

apronchenkov:
To be precise, in the case of authenticateRequest() with qop the hash check doesn't play a role. Any non-empty content leads to the 'Authenticated'.

sgodin:
I know we've been having digest auth with auth-int interop issues and I was hoping this would have fixed that, but I agree with Dario's analysis that we aren't actually calling this method anywhere.  Regardless - nice find!  Thanks.

Scott

PR #28, opened 2015-09-09
Don't use SHA-1 for SSL certificates.

fedor-brunner:
SHA-1 should not be used anymore in SSL certificates.

https://wiki.mozilla.org/CA:Problematic_Practices#SHA-1_Certificates
https://blog.mozilla.org/security/2014/09/23/phasing-out-certificates-with-sha-1-based-signature-algorithms/
http://blogs.technet.com/b/pki/archive/2013/11/12/sha1-deprecation-policy.aspx

### DIFF
--- a/resip/certs/Readme.txt
+++ b/resip/certs/Readme.txt
@@ -36,8 +36,8 @@ openssl smime -verify -in bar.msg -signer fluffy.pem -CAfile root.pem
 
 
 -- Generating a self signed cert and key -- 
-openssl genrsa -out id_key.pem 512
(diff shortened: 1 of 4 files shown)


dpocock:
The script changes appear OK

The changes to Security.cxx:

- are they being tested by any unit tests?

- why does computeIdentity still use SHA1Stream?  Is the change complete?

- what is the impact on interoperability for SIP?

- did you discuss this on any mailing list or in the bug tracker, if so, could you please provide links?

PR #29, opened 2015-09-09
use 2048-bit or stronger Diffie-Hellman groups

fedor-brunner:
reTurn was configured to use 512-bit Diffie Hellman Group.
Diffie-Hellman key exchange using such weak group can be decrypted.

https://weakdh.org/


dpocock:
This appears valid

- what is the impact on interoperability for TURN?  Will any clients be upset by this?

- did you discuss this on any mailing list or in the bug tracker, if so, could you please provide links?

fedor-brunner:
If there is a TURN client implemented in Java using Java 6, Java 7, there could be problems with 2048-bit Diffie-Hellman groups. This limitation was fixed in Java 8.

https://tt4cs.wordpress.com/2014/04/20/dh-2048-now-supported-by-jdk8/
https://stackoverflow.com/questions/10687200/java-7-and-could-not-generate-dh-keypair


A possible fix for this limitation would be to enable ECDHE key exchange in reTurn .

SamyDgz (approved):

PR #32, opened 2015-10-26
Implement server error conditions

tasifacuj:
* implement  http://tools.ietf.org/html/rfc3261#section-18.1.1, paragraph 8
Under error conditions, the
   server may attempt to open a new connection to send the response.  To
   handle this case, the transport layer MUST also be prepared to
   receive an incoming connection on the source IP address from which
   the request was sent and port number in the "sent-by" field.


sgodin:
Thanks!   This looks good to me.  I will merge shortly unless there are any objections?

tasifacuj:
Hi,
 feel free,
 thank you for the attention to this patch.

sgodin:
OK - there were no objections, code is merged.  Thank you for your contribution!

Scott

PR #33, opened 2015-10-29
Corrected call to ErrLog macro on MacOS/iOS

jasonvasquez:
There is one specific case that only occurs on iOS/MacOS (as far as I can tell), where there is a compile-time error in an attempt to log a message.  (This corrects that :))

I didn't see an immediate definition or pattern for the `error(e)` call in the surrounding code, so I am assuming at this point that the `ErrLog` call is sufficient.

### DIFF
--- a/rutil/Socket.cxx
+++ b/rutil/Socket.cxx
@@ -70,8 +70,7 @@ resip::configureConnectedSocket(Socket fd)
    if ( ::setsockopt ( fd, SOL_SOCKET, SO_NOSIGPIPE, (const char*)&on, sizeof(on)) )
    {
       int e = getErrno();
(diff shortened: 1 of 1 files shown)


sgodin:
The error method is defined as a static on the Transport class.  Transport::error - however I don't think it's really needed here.  I'm good with just removing it.  Thanks.

dpocock:
Caro amico!


Ecco un articolo che ho trovato per caso,  ma che si è scoperto così interessante che ho deciso di condividerlo  con voi, potete trovarlo qui <http://nicole.lainieofleisure.com/uwv>


In fede,Matteo Joy

SamyDgz (approved):

PR #34, opened 2015-10-29
Fix for signed types on iOS/MacOS

jasonvasquez:
On iOS/MacOS (at least with current SDKs), `MacTypes.h`  does declare `UInt8`/`UInt16`/etc for unsigned types.  But, for signed types, `MacTypes.h` defines them as `SInt8`/etc, rather than the `Int8` that the rest of the codebase expects.  This simply adds some typedefs for the signed types.

(Also, there was a minor error in the same block with a comment that accidentally was prefixed with `#`, so the preprocessor was attempting to pick it up)

### DIFF
--- a/rutil/compat.hxx
+++ b/rutil/compat.hxx
@@ -144,7 +144,7 @@ typedef char           Int8;
 typedef short          Int16;
 typedef int            Int32;
 #else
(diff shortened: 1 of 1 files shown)


sgodin:
Thanks for the contribution Jason!

PR #38, opened 2016-02-24
Ignore trailing null bytes in SIP keepalives

hypherion2:
New versions of Asterisk (12+, based on pjsip stack) send keep alive packets as '\r\n\0' instead of '\r\n\r\n'. Ignoring the fact that these are treated as "pong" packets, the trailing null byte mess up the preprocessor state such that any subsequent SIP message cannot be parsed properly and the stack sends 400 error.
This change makes the "ping" and "pong" packet processing ignore trailing null bytes.

### DIFF
--- a/resip/stack/ConnectionBase.cxx
+++ b/resip/stack/ConnectionBase.cxx
@@ -148,6 +148,12 @@ ConnectionBase::preparseNewBytes(int bytesRead)
             mBufferPos += 4;
             bytesRead -= 4;
             onDoubleCRLF();
+            // skip trailing null bytes
+            while (bytesRead && mBuffer[mBufferPos] == '\0')
+            {
+               mBufferPos++;
+               bytesRead--;
+            }
             if (bytesRead)
             {
                goto start;
@@ -165,6 +171,12 @@ ConnectionBase::preparseNewBytes(int bytesRead)
             mBufferPos += 2;
             bytesRead -= 2;
             onSingleCRLF();
+            // skip trailing null bytes
+            while (bytesRead && mBuffer[mBufferPos] == '\0')
+            {
+               mBufferPos++;
+               bytesRead--;
+            }
             if (bytesRead)
             {
                goto start;


sgodin:
We don't process Stun packets today on the TCP transports (not really sure if we ever would - RFC5626 forbids it), but the first byte of a stun request is a NULL, this logic could end up eating the first byte of a stun request.  We could always fix the logic up if we add STUN add a future date.  

I'm a little hesitant to merge this change - it seems to me the right thing to do, would be to fix Asterisks / PJSIP keepalives to not have a trailing null.  RFC5626 section 4.4.1 defines CRLF and CRLFCRLF only - no trailing nulls.  I'm also good with making us more forgiving on the receiving side, as long as we are not making compromises.   Anyone else have thoughts on this?

TafThorne:
It sounds like we can demonstrate Asterisks & PJSIP deviating from RFC5626.  I would be in favour of at least approaching those projects to see if they could explain why this was the case before altering resiprocate.  There may be some very good reasons for including the \0 which could influence the choice of how we change the resiprocate code.

hypherion2:
Issue has been reported to Asterisk JIRA: https://issues.asterisk.org/jira/browse/ASTERISK-25817.
Note that this causes frequent "random" failures when the stack happens to receive a valid SIP message after one of these keepalive packets.

dpocock:
I also feel this is not the right way to resolve the problem.  The Asterisk / PJSIP team need to adhere to specs.  If reSIProcate added non-standard behavior to cope with every broken SIP implementation out there it would become unmaintainable.  As this is only a recent regression in Asterisk, the sooner they fix it the better.  That said, if there is some particularly compelling reason we could have optional/configurable support for this, just as we have optional support for non-standard URI encoding (to cope with Cisco sending the # symbol in URIs)

dpocock:
Salve amico,

È così bello sapere che  qualcuno apprezza quello che  faccio, puoi leggere di più  sui  miei lavori qui  <http://cri.cellarsurf.com/trl>

In fede,Matteo Joy

dpocock:
Ei!

Volevo te quanto ho apprezzato molto la nostra amicizia dire,  si può leggere il  mio messaggio di ringraziamento qui <http://keep.amandalsmith.info/htk>

Spero che questo aiuti, Matteo Joy

dpocock:
@hypherion2 it looks like Digium have agreed this is an Asterisk or PJSIP bug but they haven't closed it yet.  Instead of trying to submit a workaround to reSIProcate, would you consider trying to submit a definitive fix to Asterisk?  Or even find the broken line of code in Asterisk and put a link to it in the Digium JIRA issue?

sgodin:
Closing as - not a bug.

PR #44, opened 2016-06-01
Improve logging (log errno names instead of values)

udit043:
https://project.freertc.org/issues/25

In Order to solve this bug, I have created one header file "Errdes.hxx" inside 'rutil' directory, it has function declaration and definition as well. Function on call will print error messages associated with the error number/code from the following functions : 

-> [SSL_ERROR_SYSCALL](http://www.cl.cam.ac.uk/cgi-bin/manpage?3+errno) 
-> [SSL_get_error](https://www.openssl.org/docs/manmaster/ssl/SSL_get_error.html)
-> [X509_verify_cert_error_string](https://www.openssl.org/docs/manmaster/crypto/X509_STORE_CTX_get_error.html)

I have updated all the files that were containing numeric log messages. Now these files will call function from "Errdes.hxx" and will print error message instead of error code.


sgodin:
This is great - thanks for contributing.  

One small suggestion, instead of naming the err functions generically like:  err_fun1 it would be better to use a more descriptive name:  ie:  err_fun_os.  Or better yet, since there is no real standard of using underscores in resip methods names:  
errorToStringOS
errorToStringSSL
errorToStringX509

udit043:
Okay, thanks for the suggestion. I will rename the functions name.

udit043:
Functions name has been changed.

udit043:
I am closing this pull request and will create another request on same issue with neat and clean work.

PR #45, opened 2016-06-05
buddy list and presence

MateusBellomo:
week1 (GSoC)
- support for initializing roster / buddy list
- support for adding contacts
- support for removing contacts 

week2 (GSoC)
- support for contact status changes

Still have to check why requestHandles() method is being called several times while user is typing to add a contact.


dpocock:
was this in your other pull request that we accepted?  If so, can you please close this one?

PR #49, opened 2016-06-23
log errno names instead of values

udit043:
Please visit https://github.com/udit043/resiprocate/tree/patch-6.
I have created new [branch](https://github.com/udit043/resiprocate/tree/patch-6) for this task. After discussion with mentor will close and create new pull request.

### DIFF
--- a/rutil/DnsUtil.cxx
+++ b/rutil/DnsUtil.cxx
@@ -26,6 +26,7 @@
 #include "rutil/Inserter.hxx"
 #include "rutil/WinCompat.hxx"
 #include "rutil/WinLeakCheck.hxx"
+#include "rutil/Errdes.hxx"
 
 #define RESIPROCATE_SUBSYSTEM resip::Subsystem::DNS
 
@@ -143,10 +144,10 @@ DnsUtil::getLocalHostName()
 //       current hack (see the #define in .hxx) needs
 //       to be reworked.
             case WSANOTINITIALISED:
-               CritLog( << "could not find local hostname because network not initialized:" << strerror(err) );
+               CritLog( << "could not find local hostname because network not initialized:" << strerror(err) << " error message from Errdes.hxx file : " << errortostringOS(err));
                break;
             default:
-               CritLog( << "could not find local hostname:" << strerror(err) );
+               CritLog( << "could not find local hostname:" << strerror(err) << " error message from Errdes.hxx file : " << errortostringOS(err));
                break;
          }
          throw Exception("could not find local hostname",__FILE__,__LINE__);
@@ -206,7 +207,7 @@ DnsUtil::getLocalDomainName()
          if ( e != 0 )
          {
             int err = getErrno();
-            CritLog(<< "Couldn't find domainname: " << strerror(err));
+            CritLog(<< "Couldn't find domainname: " << strerror(err)  << " error message from Errdes.hxx file : " << errortostringOS(err));
             throw Exception(strerror(err), __FILE__,__LINE__);
          }
       }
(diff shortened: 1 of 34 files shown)


rutil/DnsUtil.cxx
```diff
@@ -677,14 +678,14 @@ const char * inet_ntop6(const u_char *src, char *dst, size_t size);
 //adapted from freebsd inet_ntop.c(1.12) and inet_pton.c(1.5) for windows(non-compliant snprinf workaround)
 /* const char *
  * inet_ntop4(src, dst, size)
- *	format an IPv4 address, more or less like inet_ntoa()
+ * format an IPv4 address, more or less like inet_ntoa()
```
dpocock: please don't include these formatting changes

rutil/Errdes.hxx
```diff
@@ -0,0 +1,331 @@
...
+#include <cstring>
+
+#define MAX_ERROR_LEN 256
+static char buf[MAX_ERROR_LEN]; 
+
+/*           Below are the error description from SSL_ERROR_SYSCALL or strerror(int errno) [http://www.cl.cam.ac.uk/cgi-bin/manpage?3+errno]            */
+
+static char* messagesOS[] = {
+/*0               0   */  "No error 0",
+/*EPERM           1   */  "EPERM (Operation not permitted) 1",
```
dpocock: these numeric values are not safe to be hard-coded or used as indexes into the array

dpocock:
I've had a closer look at this

On Windows, the error names and numbers are different:
https://msdn.microsoft.com/en-us/library/windows/desktop/ms740668%28v=vs.85%29.aspx

What I'd like to see in Errdes.hxx:

* some classes:
 * a base class called ```AbstractError```
 * a subclass of ```AbstractError``` called ```NumericError```
 * three subclasses of ```NumericError``` called ```ErrnoError```, ```OpenSSLError``` and ```X509Error```

* each class should override the ```<<``` stream operator to return a string for use in log messages
* each class should provide a static factory method for getting an instance, e.g.
   ```
       const ErrnoError& ErrnoError::getErrorInstance(const int errorCode);
   ```
* instances should be cached, the getErrorInstance method should always return the same instance for a given error code
* make sure you don't assume the numeric values that the names correspond to, you must make sure you use each macro somewhere.  E.g. if you were putting them in a hash map at initialization time, you might do:
   ```
      my_map[ENOENT] = ErrnoError(ENOENT, "ENOENT", "No such file or directory");
   ```
* use the relevant object in the code, e.g. wherever there is a call to getErrno() or strerror(), use an instance of ErrnoError instead.

udit043:
Ok, working on it.

dpocock:
for now, just create the classes, don't worry about changing all the other code to use them yet.  Let me look at the classes you create and we'll refine them before you change all the other code.

udit043:
Here are the links to the new error files:
Errdes.hxx : https://github.com/udit043/resiprocate/blob/patch-2/rutil/Errdes.hxx
Errdes.cxx : https://github.com/udit043/resiprocate/blob/patch-2/rutil/Errdes.cxx
Way the files using the error file : 
1 - https://github.com/resiprocate/resiprocate/commit/ea5b5190ebdcd48b4d5cce14be73f092d0f5171a
2 - https://github.com/resiprocate/resiprocate/commit/5179e3fb37cf6c542cf6de0375d02ee47756e2ba

udit043:
Please visit https://github.com/udit043/resiprocate/tree/patch-6
I have created new [branch](https://github.com/udit043/resiprocate/tree/patch-6) and implement class and mapping for this work. After discussion with mentor will close and create new pull request.

dpocock:
Udit, can you please create a new pull request for me to comment on?

PR #50, opened 2016-07-10
bug fix for methods overriding onSubscriptionNotify of UserAgent

MateusBellomo:
the methods signature must be exactly like as resip/recon/UserAgent.* onSubscriptionNotify()


dpocock:
thanks for this contribution

PR #53, opened 2016-07-29
send/receive text message mechanism through MESSAGE message

MateusBellomo:
changes at resip/recon API to enable send and receive MESSAGE messages.

### DIFF
--- a/resip/recon/UserAgent.cxx
+++ b/resip/recon/UserAgent.cxx
@@ -21,6 +21,9 @@
 #include <resip/dum/ClientSubscription.hxx>
 #include <resip/dum/ServerSubscription.hxx>
 #include <resip/dum/ClientRegistration.hxx>
+#include <resip/dum/ClientPagerMessage.hxx>
+#include <resip/dum/ServerPagerMessage.hxx>
+#include <resip/stack/PlainContents.hxx>
 #include <resip/dum/KeepAliveManager.hxx>
 #include <resip/dum/AppDialogSet.hxx>
 #if defined(USE_SSL)
@@ -88,6 +91,10 @@ UserAgent::UserAgent(ConversationManager* conversationManager, SharedPtr<UserAge
    mDum.addOutOfDialogHandler(REFER, mConversationManager);
    mDum.addClientSubscriptionHandler("refer", mConversationManager);
    mDum.addServerSubscriptionHandler("refer", mConversationManager);
+
+   InstantMessage* mInstantMessage = new recon::InstantMessage();
+   mDum.setServerPagerMessageHandler(mInstantMessage);
+   mDum.setClientPagerMessageHandler(mInstantMessage);
 
    //mDum.addClientSubscriptionHandler(Symbols::Presence, this);
    //mDum.addClientPublicationHandler(Symbols::Presence, this);
@@ -450,6 +457,24 @@ void
 UserAgent::onSubscriptionNotify(SubscriptionHandle handle, const Data& notifyData)
 {
    // Default implementation is to do nothing - application should override this
+}
+
+const char*
+UserAgent::sendMessage(const NameAddr& destination, const Data& msg, const Mime& mimeType)
+{
+   if(!mDum.getMasterProfile()->isMethodSupported(MESSAGE))
+   {
+      WarningLog (<< "MESSAGE method not detected in list of supported methods, adding it belatedly" );
+      mDum.getMasterProfile()->addSupportedMethod(MESSAGE);
+   }
+   
+   ClientPagerMessageHandle cpmh = mDum.makePagerMessage(destination);
+   auto_ptr<Contents> msgContent(new PlainContents(msg, mimeType));
+   cpmh.get()->page(msgContent);
+   SharedPtr<SipMessage> sipMessage = cpmh.get()->getMessageRequestSharedPtr();
+   mDum.send(sipMessage);
+
+   return sipMessage->header(h_CallId).value().c_str();
 }
 
 void 
--- a/resip/recon/InstantMessage.cxx
+++ b/resip/recon/InstantMessage.cxx
@@ -1,0 +1,84 @@
+#include "ReconSubsystem.hxx"
+#include "InstantMessage.hxx"
+
+#include <rutil/Log.hxx>
+#include <rutil/Logger.hxx>
+
+#include <resip/dum/ClientPagerMessage.hxx>
+#include <resip/dum/ServerPagerMessage.hxx>
+#include <resip/stack/PlainContents.hxx>
+
+#include "ReconSubsystem.hxx"
+#include "UserAgent.hxx"
+
+using namespace recon;
+using namespace resip;
+using namespace std;
+
+#define RESIPROCATE_SUBSYSTEM ReconSubsystem::RECON
+
+InstantMessage::InstantMessage()
+{
+
+}
+
+InstantMessage::~InstantMessage()
+{
+
+}
+
+void
+InstantMessage::onMessageArrived(resip::ServerPagerMessageHandle handle, const resip::SipMessage& message)
+{
+   // Default implementation is to do nothing - application should override this
+}
+
+void
+InstantMessage::onSuccess(resip::ClientPagerMessageHandle handle, const resip::SipMessage& status)
+{
+   // Default implementation is to do nothing - application should override this
+}
+
+void
+InstantMessage::onFailure(resip::ClientPagerMessageHandle handle, const resip::SipMessage& status, std::auto_ptr<resip::Contents> contents)
+{
+   // Default implementation is to do nothing - application should override this
+}
+
+/* ====================================================================
+ *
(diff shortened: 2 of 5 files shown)


resip/recon/ConversationManager.cxx
```diff
@@ -1223,6 +1226,30 @@ ConversationManager::onTryingNextTarget(AppDialogSetHandle, const SipMessage& ms
...
+const char*
+ConversationManager::sendMessage(const NameAddr& destination, const Data& msg, const Mime& mimeType)
+{
+   if(!mUserAgent->mDum.getMasterProfile()->isMethodSupported(MESSAGE))
+   {
+      mUserAgent->mDum.getMasterProfile()->addSupportedMethod(MESSAGE);
+   }
+   if(!mUserAgent->mDum.getMasterProfile()->isMimeTypeSupported(MESSAGE, mimeType))
+   {
+      mUserAgent->mDum.getMasterProfile()->addSupportedMimeType(MESSAGE, mimeType);
```
dpocock: we shouldn't do this here in recon, the application developer needs to decide which MIME types they want to support so please remove this

resip/recon/ConversationManager.cxx
```diff
@@ -1223,6 +1226,30 @@ ConversationManager::onTryingNextTarget(AppDialogSetHandle, const SipMessage& ms
...
+   }
+   
+   ClientPagerMessageHandle cpmh = mUserAgent->mDum.makePagerMessage(destination);
+   auto_ptr<Contents> msgContent(new PlainContents(msg));
```
dpocock: please use the PlainContents constructor that also takes the MIME type as a second argument

resip/recon/ConversationManager.hxx
```diff
@@ -483,6 +486,18 @@ public:
    virtual void onDtmfEvent(ParticipantHandle partHandle, int dtmf, int duration, bool up) = 0;
 
    ///////////////////////////////////////////////////////////////////////
+   // Pager Message Handlers /////////////////////////////////////////////
+   ///////////////////////////////////////////////////////////////////////
+
+   virtual void onMessageArrived(resip::ServerPagerMessageHandle handle, const resip::SipMessage& message) = 0;
```
dpocock: how does this relate to ConversationManager::onMessage(resip::InviteSessionHandle, const resip::SipMessage& msg) ?

resip/recon/ConversationManager.hxx
```diff
@@ -483,6 +486,18 @@ public:
...
+   // Called when a MESSAGE has been successfully sent
+   virtual void onSuccess(resip::ClientPagerMessageHandle handle, const resip::SipMessage& status) = 0;
+   // Application could re-page the failed contents or just ingore it.
+   virtual void onFailure(resip::ClientPagerMessageHandle handle, const resip::SipMessage& status, std::auto_ptr<resip::Contents> contents) = 0;
+   
+   const char* sendMessage(const resip::NameAddr& destination, const resip::Data& msg, const resip::Mime& mimeType);
```
dpocock: this is a method that may belong in UserAgent instead of ConversationManager because MESSAGE doesn't create a dialog.  It is discussed in RFC 3428 section 2
https://tools.ietf.org/html/rfc3428#section-2

dpocock: then again, it depends on how we define "Conversation", even though there is no SIP dialog we could consider a series of messages to be a Conversation

resip/recon/UserAgent.cxx
```diff
@@ -89,6 +89,9 @@ UserAgent::UserAgent(ConversationManager* conversationManager, SharedPtr<UserAge
    mDum.addClientSubscriptionHandler("refer", mConversationManager);
    mDum.addServerSubscriptionHandler("refer", mConversationManager);
 
+   mDum.setServerPagerMessageHandler(mConversationManager);
+   mDum.setClientPagerMessageHandler(mConversationManager);
```
dpocock: instead of using mConversationManager here, it could be some other object for managing IM sessions

resip/recon/ConversationManager.hxx
sgodin: Client and Server Pager Message handler classes are intended to be used for out of dialog MESSAGE requests.

resip/recon/ConversationManager.hxx
sgodin: It would have been much nicer (and more work) to have use the concept of participants and allow message participants in a conversation, instead of this quick an dirty - here's a message, you figure out what to do with it.  If we allowed conversations with multiple message participants we could implement a chat group concept in recon.

resip/recon/ConversationManager.hxx
MateusBellomo: The method onMessageArrived() is called at [1] using the ServerPagerMessageHandler that was set at resip/recon/UserAgent.cxx:

mDum.setServerPagerMessageHandler(mConversationManager);

The method onMessage() is coming from InviteSessionHandler that ConversationManager derived from. Do you think I could use this same method for incoming text messages?

[1] https://github.com/resiprocate/resiprocate/blob/master/resip/dum/ServerPagerMessage.cxx#L86

resip/recon/ConversationManager.hxx
MateusBellomo: Do you think it's better to create another class at resip/recon that handles all the text message stuff? Then I could make this new class extend the ServerPagerMessageHandler and  ClientPagerMessageHandler as I'm doing with resip/recon/ConversationManager now.

resip/recon/ConversationManager.hxx
sgodin: The InviteSessionHandler onMessage is for in-dialog MESSAGE requests only
(ie:  MESSAGE requests received in the context of an existing INVITE
session).  The standard for IM's is to use out-of-dialog MESSAGE requests,
so you may want to handle both, but definitely want to handle at least
using the Pager classes.

Scott

MateusBellomo: The out-of-dialog MESSAGE request is already handle when I made ConversationManager derives from ServerPagerMessageHandler and ClientPagerMessageHandler and implement it's methods, right?
Is there an example on how to handle in-dialog MESSAGE requests and how to create a dialog from a out-of-dialog MESSAGE?

resip/recon/ConversationManager.hxx
dpocock: @sgodin do you mean the existing ```Conversation``` object though, or should some other ```Conversation``` object be used for chat?

E.g. we could have an abstract ```Conversation``` and then subclasses ```ChatConversation``` and ```MediaConversation```

The distinction between the two may be the existence of a SIP dialog.

Given that conversations can change (e.g. they start on chat, then do some webcam call, then fall back to chat for a while), maybe it is better not to have such a class hierarchy though.

When recon is used on the desktop (which is the aim here), the user may have a chat conversation with one peer and a combined chat/webcam session with another peer concurrently, but those would be 2 or 3 different Conversation instances.

If you are not happy for this to go into the API, what would you feel is a reasonable bare-minimum implementation of this that we could accept in recon for now?  I can also spend some time helping on the API side over the next few days while Mateus finishes the telepathy component.

sgodin: Off the top of my head,  I think having out-of-dialog IM's and audio/video
in a single conversation is valuable.  I like this better than having two
different conversation classes.  We could use the From and/or To headers to
correlate out-of-dialog IMs with existing audio or video calls.  Using your
desktop client example, it would be ideal if there was only one
conversation for all media types.

This doesn't seem like something that could be done in a few weeks
time-frame though.

I don't think the current implementation is ideal - but I also don't think
the it prevents us from doing it differently in the future (unless you care
about upgrade support).

Scott

resip/recon/ConversationManager.hxx
sgodin: Yes that's correct.  Handling in-dialog MESSAGE requests is simply handling
the onMessage API from InviteSessionHandler.  You don't create a dialog
specifically for MESSAGE handling - but you could send an in-dialog MESSAGE
on and existing call using the InviteSession::message API.  I'm not sure
you should worry about in-dialog MESSAGEs at this point, if you don't have
a direct need for them.

resip/recon/UserAgent.cxx
```diff
@@ -452,6 +459,23 @@ UserAgent::onSubscriptionNotify(SubscriptionHandle handle, const Data& notifyDat
    // Default implementation is to do nothing - application should override this
 }
 
+const char*
+UserAgent::sendMessage(const NameAddr& destination, const Data& msg, const Mime& mimeType)
+{
+   if(!mDum.getMasterProfile()->isMethodSupported(MESSAGE))
+   {
```
dpocock: if MESSAGE wasn't already in the MasterProfile, we wouldn't be able to receive MESSAGE requests.  Please log a warning message here when this scenario is detected.

MateusBellomo: but this function is for sending messages, it would get here when a MESSAGE is received?

PR #54, opened 2016-07-30
Client support for TLS SNI extension

udit043:
Server Name Indication (SNI) is an extension to the TLS networking protocol by which a client indicates which hostname it is attempting to connect to at the start of the handshaking process.
In this pull request I am enabling client support for TLS SNI extension.
Here are some screenshots:
Client Hello without SNI supported: https://i.imgsafe.org/54bf8257b4.png 
Client Hello with SNI supported: https://i.imgsafe.org/54c253a037.png

### DIFF
--- a/resip/stack/ssl/TlsConnection.cxx
+++ b/resip/stack/ssl/TlsConnection.cxx
@@ -206,6 +206,11 @@ TlsConnection::checkState()
       else
       {
          InfoLog( << "TLS handshake starting (client mode)" );
+         /* OpenSSL version < 0.9.8f does not support SSL_set_tlsext_host_name() */
+         #if defined(SSL_set_tlsext_host_name)
+            SSL_set_tlsext_host_name(mSsl,who().getTargetDomain().c_str()); // set the SNI hostname
+            DebugLog ( << "TLS SNI extension in Client Hello: " << who().getTargetDomain().c_str());
+         #endif
          SSL_set_connect_state(mSsl);
          mTlsState = Handshaking;
       }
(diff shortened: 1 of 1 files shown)


dpocock:
Discussed on the mailing list

http://list.resiprocate.org/archive/resiprocate-devel/msg09220.html

resip/stack/ssl/TlsConnection.cxx
```diff
@@ -206,6 +206,11 @@ TlsConnection::checkState()
       else
       {
          InfoLog( << "TLS handshake starting (client mode)" );
+         /* OpenSSL version < 0.9.8f does not support SSL_set_tlsext_host_name() */
+         #if defined(SSL_set_tlsext_host_name)
```
dpocock: please move the macro lines (#if and #endif) all the way to the left, no indentation

move the code 3 spaces to the left, so it is aligned with the rest of the code in this block

resip/stack/ssl/TlsConnection.cxx
```diff
@@ -206,6 +206,11 @@ TlsConnection::checkState()
...
+            SSL_set_tlsext_host_name(mSsl,who().getTargetDomain().c_str()); // set the SNI hostname
+            DebugLog ( << "TLS SNI extension in Client Hello: " << who().getTargetDomain().c_str());
```
dpocock: put the log line above the SSL_set_tlsext_host_name line

You don't need to use c_str() in the log line

shubhamfuloria:
```cpp
DebugLog ( << "TLS SNI extension in Client Hello: " << who().getTargetDomain());SSL_set_tlsext_host_name(mSsl,who().getTargetDomain().c_str()); // set the SNI hostname
```

Why does who().getTargetDomain().c_str() returns ip address of client, instead of domain name ? Do I need to set domain name somewhere in the application ?

PR #56, opened 2016-08-18
load/save buddies list, send/receive presence, send/receive text messages

MateusBellomo:
These changes enables a contact list, the possibility of seeing the contacts status, sending the user status to contacts, and send/receive text messages.

### DIFF
--- a/apps/telepathy/Connection.cxx
+++ b/apps/telepathy/Connection.cxx
@@ -29,6 +29,7 @@
 #include "Connection.hxx"
 #include "Common.hxx"
 #include "SipCallChannel.hxx"
+#include "TextChannel.hxx"
 
 using namespace recon;
 using namespace resip;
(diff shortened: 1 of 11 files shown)


apps/telepathy/Connection.cxx
```diff
@@ -142,6 +152,198 @@ tr::Connection::Connection(const QDBusConnection &dbusConnection, const QString
...
+      return;
+   }
+   
+   qDebug() << Q_FUNC_INFO << aliases;
+   
+   for ( Tp::AliasMap::const_iterator it = aliases.begin(); it != aliases.end(); it++ )
+   {
+      mAliases[it.key()] = it.value();
+   }
+
```
Kaffeine: mAliasingInterface->aliasesChanged(); call is missing here.

PR #58, opened 2016-08-21
update gloox parameters from v0.9.9.8 to v1.0.16

balumenon96:
This patch upgrades support to the latest build of gloox library to-date, v1.0.16. 
To fix, registerIqHandler warning for apple::iq variables. This does not effect the working on other systems.

### DIFF
--- a/apps/ichat-gw/jabberconnector/JabberComponent.cxx
+++ b/apps/ichat-gw/jabberconnector/JabberComponent.cxx
@@ -13,9 +13,11 @@
 #ifndef RESIP_CONTRIB_GLOOX
 #include <gloox/disco.h>
 #include <gloox/mutex.h>
+#include <gloox/message.h>
 #else
 #include <src/disco.h>
 #include <src/mutex.h>
+#include <src/message.h>
 #endif
 
 using namespace gateway;
@@ -119,7 +121,7 @@ public:
    virtual void lock() { mMutex.lock(); }
    virtual void unlock() { mMutex.unlock(); }
 private:
-   gloox::Mutex mMutex;
+   gloox::util::Mutex mMutex;
 };
 IPCMutexGloox g_IPCGlooxMutex;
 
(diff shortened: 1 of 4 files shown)


apps/ichat-gw/jabberconnector/JabberComponent.cxx
```diff
@@ -966,25 +968,25 @@ JabberComponent::handlePresence(Stanza *stanza)
...
+JabberComponent::handleMessage(const Message& msg, MessageSession* session)
 {
    std::ostringstream oss;
-   oss << "JabberComponent::handlePresence - " << *stanza;
+   oss << "JabberComponent::handlePresence - " << &msg;
    handleLog(gloox::LogLevelDebug, gloox::LogAreaUser, oss.str());
-   Stanza *s = Stanza::createMessageStanza(stanza->from().full(), "You have reached the ichat gateway!" );
-   s->addAttribute("from", stanza->to().full());
+   Message *s = new Message(Message::Normal, msg.from().full(), "You have reached the ichat gateway!" );
+   s->tag()->addAttribute("from", msg.to().full());
```
dpocock: I wonder if you need to do anything else when constructing a message to avoid that error in prosody.log

Can you find any other examples of code based on gloox and see how they construct a Message?

balumenon96: This upgrade was intended to update the existing functions to the latest version. It is then to be tested and further changes will be made on this code, but as a separate pull request.

dpocock:
Balram, it looks like you need to stop it including the xmlns attribute:
https://mail.jabber.org/pipermail/jdev/2016-August/090263.html

Check the gloox API, maybe there is some setting to distinguish when it behaves as a client and when it is used as a component.

dpocock:
@sgodin would you like to make any comments on Balram's work or do you know anybody else who should comment on it?

sgodin:
Just took a quick look - seems like adjustments for new gloox version
only.  I don't see any immediate issues, and unfortunately I don't have
time to review in any more detail.

Scott

SamyDgz (approved):

PR #61, opened 2016-08-24
fixed PUBLISH pidf order and attributes

MateusBellomo:


resip/recon/UserAgent.cxx
```diff
@@ -649,13 +649,28 @@ UserAgent::createPublicationImpl(PublicationHandle handle, const Data& status, c
...
+
    // Adding rpid defined at RFC 4480 https://tools.ietf.org/html/rfc4480#page-7
    resip::GenericPidfContents gPidf;
+   gPidf.setSimplePresenceTupleNode(Random::getRandomHex(3), true, Data::Empty, note, target.uri().getAorNoReally(), Data(1.0));
 
    resip::GenericPidfContents::Node* dm = new resip::GenericPidfContents::Node();
    dm->mNamespacePrefix = Data("dm:");
    dm->mTag = "person";
-   dm->mAttributes["id"] = Random::getRandomHex(3);
+   dm->mAttributes["id"] = Data(string("ID-") + Random::getRandomHex(3).c_str());
```
saghul: you probably want to do the same for the tuple ID since it's of the same type (xs:ID). It cannot begin with a number, so a predefined prefix (such as "ID-") works.

MateusBellomo: done!

PR #74, opened 2017-04-11
MessageWaitingContents: skip body lines which do not match rfc 3842.

jhcloos:
If the line matches /^[fmnptv/i but does not match /^\w+-Message:/
it is skiped.

This ignores lines like /^Voicemail: \d+/\d+$/ from before rfc 3842
which some software still send for backward compatibility.

Signed-off-by: James Cloos <cloos@jhcloos.com>

### DIFF
--- a/resip/stack/MessageWaitingContents.cxx
+++ b/resip/stack/MessageWaitingContents.cxx
@@ -361,6 +361,14 @@ MessageWaitingContents::parse(ParseBuffer& pb)
       }
       resip_assert(ht != -1);
 
+      const char* bol = pb.position();
+      pb.skipToChars("essage");
+      if (pb.position() == bol)
+      {
+	pb.skipChars(Symbols::CRLF);
+	continue;
+      }
+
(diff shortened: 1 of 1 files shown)


sgodin:
Hmmm, this isn't quite what I was picturing when I suggested it.  I thought you were suggesting to look for the entire header string (case insensitive) instead of just the first letter for all header types.

I'm not sure I like the solution you presented.  It took me a while to figure out exactly what you were doing here, and the fact that all the headers actually contain the word Message somewhere in them.  With the exception of the none type - that I don't believe will work at all now.

Can you change this to tokenize the header name first, then do a case insensitve compare (on the full header text) for each accepted type to assign the ht (header type) to it?  If the header token is not recognized, then we will can happily ignore it.

jhcloos:
Sure.

I kept the current initial letter check presuming that a minimal change
would be preferred.

But I can switch it to a full compare.

-JimC
-- 
James Cloos <cloos@jhcloos.com>         OpenPGP: 0x997A9F17ED7DAEA6

TheGiraffe3:
Is this going to happen?

PR #75, opened 2017-05-23
the SHA1 hash needs to be 5 chunks 32 bits each

marcb1:
unsigned long int is 64 bits on a 64 bit machine. This was resulting in an incorrect SHA1 hash.


dpocock:
@marcb1 did you close this because it is not an issue or because the fix wasn't correct?

Do you think you could add a unit test to prove this fails on 64 bit at present?

PR #77, opened 2017-06-26
Add ares fuzzer

gjasny:


gjasny:
I have another patch to fix an heap overflow in ares_mkquery but would like to give it some more testing.

gjasny:
Could you please consider merging? All make targets are optional and nut build by default.

dpocock:
Thanks for this contribution.  I'm using c-ares (the package on Linux distributions) so I haven't tested this, it may better for @sgodin to decide on changes to ares code as I think he uses this in Windows builds.

sgodin:
Hi Daniel,

I merged the ares fix itself.  The other Pull request is touches the linux
build system, so I thought you were best to review that.  I'm OK with test
source files in the request, they will not effect the stack in anyway.

Gregor what is the purpose of these "fuzzer" files?

Thanks for the contribution!
Scott

gjasny:
It allows us to feed the ares DNS parser code with random junk and detect buffer overflows and memory leaks. We could do the same with the SIP or SDP parser.

Google provides an service to run and report those tests automatically: https://github.com/google/oss-fuzz/blob/master/README.md

PR #82, opened 2017-07-12
Add rutil and resip/stack fuzzers

gjasny:


sgodin:
Hi Gregor,

Please try to stick to resip bracketing style with curly braces on their own line (see/fixup fuzzStack.cxx).  Should fuzzStack be called fuzzSdp instead, or are you envisioning other fuzz tests for the stack to be added here over time?  

I'm good with the changes, but will let Daniel have a look since there are Makefile changes.

Thanks!
Scott

gjasny:
Hi Scott,

I intend to add more tests around all the parsers. I have the next one on the queue but need to fix the URI parser first.

Thanks,
Gregor

gjasny:
I split the functional change in the Data class into #90.

PR #83, opened 2017-07-13
Add OpenSSL 1.1 compatibility

gjasny:
Hello,

I tried to migrate reSIProcate to the OpenSSL 1.1 API. Those changes have been compile tested with OpenSSL 1.0.2 and OpenSSL 1.1.0.

Unfortunately my (D)TLS and OpenSSL knowledge is quite small. So I consider this PR just a discussion and review starter instead of a polished PR.

Thanks,
Gregor

### DIFF
--- a/reflow/dtls_wrapper/bf_dwrap.c
+++ b/reflow/dtls_wrapper/bf_dwrap.c
@@ -7,8 +7,37 @@
 #include <stdio.h>
 #include <errno.h>
 #include <openssl/bio.h>
+#include <openssl/opensslv.h>
 #include "rutil/ResipAssert.h"
 #include <memory.h>
+
+#if OPENSSL_VERSION_NUMBER < 0x10100000L
+
+static inline BIO_METHOD *BIO_meth_new(int type, const char *name)
+{
+  BIO_METHOD *biom = calloc(1, sizeof(BIO_METHOD));
+
+  if (biom != NULL) {
+    biom->type = type;
+    biom->name = name;
+  }
+  return biom;
+}
+
+#define BIO_meth_set_write(b, f) (b)->bwrite = (f)
+#define BIO_meth_set_read(b, f) (b)->bread = (f)
+#define BIO_meth_set_puts(b, f) (b)->bputs = (f)
+#define BIO_meth_set_gets(b, f) (b)->bgets = (f)
+#define BIO_meth_set_ctrl(b, f) (b)->ctrl = (f)
+#define BIO_meth_set_create(b, f) (b)->create = (f)
+#define BIO_meth_set_destroy(b, f) (b)->destroy = (f)
+#define BIO_meth_set_callback_ctrl(b, f) (b)->callback_ctrl = (f)
+
+#define BIO_set_init(b, val) (b)->init = (val)
+#define BIO_set_data(b, val) (b)->ptr = (val)
+#define BIO_get_data(b) (b)->ptr
+
+#endif /* OPENSSL_VERSION_NUMBER < 0x10100000L */
 
 #define BIO_TYPE_DWRAP       (50 | 0x0400 | 0x0200)
 
--- a/reflow/dtls_wrapper/bf_dwrap.c
+++ b/reflow/dtls_wrapper/bf_dwrap.c
@@ -86,9 +86,12 @@ static int dwrap_new(BIO *bi)
 
 static int dwrap_free(BIO *a)
 {
-   if(a) return 0;
+   if(!a) return 0;
 
-   free(a);
+   OPENSSL_free(BIO_get_data(a));
+   BIO_set_data(a, NULL);
+   BIO_set_init(a, 0);
+
    return 1;
 }
 
--- a/reflow/dtls_wrapper/bf_dwrap.c
+++ b/reflow/dtls_wrapper/bf_dwrap.c
@@ -21,20 +21,6 @@ static int dwrap_gets(BIO *b, char *buf, int size);
 static long dwrap_ctrl(BIO *b, int cmd, long num, void *ptr);
 static long dwrap_callback_ctrl(BIO *b, int cmd, bio_info_cb *fp);
 
-static BIO_METHOD methods_dwrap=
-{
-   BIO_TYPE_DWRAP,
-      "dtls_wrapper",
-      dwrap_write,
-      dwrap_read,
-      dwrap_puts,
-      dwrap_gets,
-      dwrap_ctrl,
-      dwrap_new,
-      dwrap_free,
-      dwrap_callback_ctrl
-};
-
 typedef struct BIO_F_DWRAP_CTX_ 
 {
    int dgram_timer_exp;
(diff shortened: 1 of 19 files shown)


resip/stack/ssl/DtlsTransport.cxx
```diff
@@ -227,14 +243,13 @@ DtlsTransport::_read( FdSet& fdset )
...
-   ssl->rbio = rbio ;
+   SSL_set0_rbio( ssl, rbio );
 
    len = SSL_read( ssl, pt, UdpTransport::MaxBufferSize ) ;
    int err = SSL_get_error( ssl, len ) ;
 
    /* done with the rbio */
-   BIO_free( ssl->rbio ) ;
-   ssl->rbio = mDummyBio ;
+   SSL_set0_rbio( ssl, mDummyBio );
```
gjasny: I'm not sure if this could lead to errors. Formerly `BIO_free` was used. Now `BIO_free_all` is used instead. IMHO the difference is that the later is freeing recursively. But our `rbio` seems to be not stacked, so the difference might not matter at all.

gjasny:
In the debian bug tracker is also a patch: https://bugs.debian.org/cgi-bin/bugreport.cgi?att=1;bug=859715;filename=resip.patch;msg=15

gjasny:
Hey @FauxFaux, I merged my patch (which preserves OpenSSL 1.0 compatibility) with yours. Could you please give it a review?

Differences from your patch to this PR are: https://gist.github.com/gjasny/04463e481d9b14e81c08d0fa3b9071af

resip/stack/ssl/Security.hxx
```diff
@@ -181,7 +181,7 @@ class BaseSecurity
       // retrieves a list of all certificate names (subjectAltNAme's and CommonName)
       static void getCertNames(X509 *cert, std::list<PeerName> &peerNames, bool useEmailAsSIP = false);
 
-      static bool isSelfSigned(const X509* cert);
+      static bool isSelfSigned(X509* cert);
```
Lastique: This should not be needed. The operation is not supposed to modify the certificate and both `X509_get_issuer_name` and `X509_get_subject_name` accept `const X509*`.

gjasny: In OpenSSL 1.0.2 the property accessors just take a non-const `X509` pointer. 
See [x509.h](https://github.com/openssl/openssl/blob/OpenSSL_1_0_2-stable/crypto/x509/x509.h#L905)

Lastique: Then, I believe, `isSelfSigned` should perform the `const_cast` internally or use the old code for OpenSSL 1.0.x. There is no reason to change the public interface of `isSelfSigned`.

reflow/dtls_wrapper/bf_dwrap.c
```diff
@@ -43,19 +58,28 @@ typedef struct BIO_F_DWRAP_CTX_
 
 BIO_METHOD *BIO_f_dwrap(void) 
 {
-   return(&methods_dwrap);
+   BIO_METHOD *meth = BIO_meth_new(BIO_TYPE_DWRAP, "dtls_wrapper");
```
Lastique: I believe, this memory leaks.

Lastique:
Also, I believe, `dwrap_free` in `reflow/dtls_wrapper/bf_dwrap.c` needs to be updated. It should free the `BIO_F_DWRAP_CTX` structure allocated in `dwrap_new`, not the `BIO` object itself.

Lastique:
In `rutil/ssl/OpenSSLInit.cxx`, `OpenSSLInit::~OpenSSLInit()`, the call `sk_SSL_COMP_free( SSL_COMP_get_compression_methods())` causes a crash with OpenSSL 1.1.0 because libssl does free the compression methods itself when being unloaded. This line is viable for OpenSSL until 1.0.2, which introduced `SSL_COMP_free_compression_methods` for this purpose. In OpenSSL 1.1.0 this call is not needed and should be removed.

resip/stack/ssl/Security.hxx
gjasny: Fixed.

reflow/dtls_wrapper/bf_dwrap.c
gjasny: I will try to create an unit test and check for memory leaks.

gjasny:
@Lastique I think I addressed all of your comments. The dtls_wrapper tests no exit without leaked memory. (For enabling and fixing those tests see future PR).

The `sk_SSL_COMP_free` code seems to be commented out.

reflow/dtls_wrapper/bf_dwrap.c
```diff
@@ -21,50 +56,60 @@ static int dwrap_gets(BIO *b, char *buf, int size);
...
 {
-   return(&methods_dwrap);
+   BIO_meth_free(dtls_meth);
+   dtls_meth = NULL;
+}
+
+BIO_METHOD *BIO_f_dwrap(void)
+{
+   if (!dtls_meth) {
+      dtls_meth = BIO_meth_new(BIO_TYPE_DWRAP, "dtls_wrapper");
```
Lastique: This is not thread-safe. Not sure if there are cases when this code can run in different threads, though. I would recommend using `pthread_once` or atomics instead.

gjasny: You're right but this codebase is still C++03. How about using a static `auto_ptr` holding the `BIO_method`? That would also enable me to get rid of the `atexit` hack.

Lastique: A simpler solution is to convert `dtls_meth` to an object with a constructor and destructor and move the `BIO_METHOD` initialization to the constructor. This would still be not thread-safe, but in practical sense this would be better because the constructor will be run at library load time, when there are typically no multiple threads.

gjasny: OTOH who knows if calling into OpenSSL works before initalization ...

Lastique: > How about using a static auto_ptr holding the BIO_method?

This won't help. The race is between loading the pointer for testing against `NULL` and storing the allocated pointer. `auto_ptr` doesn't help with it.

Re atomics, I was meaning compiler-specific intrinsics. One would likely have to introduce some abstraction layer for portability. `pthread_once` is available for C++03 programs but not on Windows. I would probably use `Interlocked*` intrinsics on Windows and `pthread_once` everywhere else.

gjasny: Will work on something like this. But I have to finish some unrelated work first.

Lastique: > OTOH who knows if calling into OpenSSL works before initalization

I would assume it should work for the methods we need - those are just memory allocation and accessors, no apparent reason for them to not work.

gjasny: @Lastique I changed the code as you suggested. Could you pleas give it another review?
 
Thanks a lot for your time!

gjasny:
@sgodin Are you ok with merging this?

sgodin:
Hi Gregor,

I haven't had a chance to try it out, but I trust you have, and I see a
number of names on this pull request that have, so I'm good with it.  If
you haven't tried a Windows build and there are issues after you merge, I
can try to resolve those ASAP.

Thanks for your efforts with this!!

Scott

PR #94, opened 2017-10-19
Create repro.service

nshopik:
add systemd unit file

We found that repro crash from time to time due bug in handling some unknown to us sip packets so we come up with simple systemd unit file to restart repro after crash

### DIFF
--- a/debian/repro.service
+++ b/debian/repro.service
@@ -1,0 +1,12 @@
+[Unit]
+Description=reSIProcate SIP stack - lightweight SIP proxy daemon
+After=local-fs.target network.target
+
+[Service]
+Type=simple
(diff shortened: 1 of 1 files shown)


nshopik:
@sgodin can you merge it? could be useful in future debian package

gjasny:
I'll take care of this PR within the next day(s)

gjasny:
Thanks for bringing up systemd support. There are already service files within the code base:

```
reTurn/pkg/common/resiprocate-turn-server.service
apps/telepathy/resiprocate.service
repro/pkg/common/repro.service
```

To install them properly I synced (#97) the Debian packaging scripts back from latest Sid into reSIProcate. I tested repro and it works as expected.

PR #100, opened 2018-01-09
fix a sip message body bug

arrayt:
This fixes a sip message body issue when get the message from TCP connection. If content length is set to 0, the message body should not be set, otherwise it is considered has body but actually has not, and result to incorrect status UAS_Offer in ServerInviteSession, but it should be UAS_NoOffer.

Please refer code diff, and let me know if any question.

### DIFF
--- a/resip/stack/ConnectionBase.cxx
+++ b/resip/stack/ConnectionBase.cxx
@@ -403,7 +403,8 @@ ConnectionBase::preparseNewBytes(int bytesRead)
                }
 
                // The message body is complete.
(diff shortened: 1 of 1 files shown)


sgodin:
Thanks for the contribution.  Just a couple things:
1.  resip coding style is to use curly braces for all statements - I am aware not all current code follows this.  :)
2.  I'm curious exactly what triggers this problem.  I've send and received many SIP message over TCP without a SIP body and there are no problems.  Any idea why you are seeing this and others are not?

PR #101, opened 2018-02-18
dtls_wrapper: Don't use std namespace

gjasny:


sgodin:
Hi Gregor,

I have no issues with this change, just curious as to why it was required.  I'm assuming you had a naming collision with something else?

Scott

gjasny:
In C++11 mode I get a conflict between `std::bind` and `bind`. I think `using namespace std` is discouraged even in cpp files. I could also qualify the bind call.

sgodin:
what you did was fine, I was just curious why it was required.  I added your note about std::bind and bind naming collision on the merge notes.. Thanks!

PR #103, opened 2018-02-18
rutil: Remove unreferenced DigestStream

gjasny:
It is unreferenced, Windows ignores it, autotools just ships the header.


sgodin:
This could be used by someone, even if the stack isn't using it.  Is there a reason to delete the files?

gjasny:
The implementation isn't included in any of the libs, so the header is of little use. I could also re-add it and fix the code / tests.

sgodin:
just hate to delete potentially good code, that might be of use to someone in the future.  If the code doesn't work, then I agree, let's delete it.

PR #108, opened 2018-03-05
rutil: ParseBuffer: Do not rely on undefined behavior

gjasny:
The previous overflow detection relied on undefined behavior
because a signed integer overflow is undefined. Compilers
are allowed to remove the overflow check completely.

That issue was found by the oss-fuzz project.

Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=6712
Bug: https://bugs.chromium.org/p/oss-fuzz/issues/detail?id=6717

### DIFF
--- a/rutil/ParseBuffer.cxx
+++ b/rutil/ParseBuffer.cxx
@@ -656,10 +656,10 @@ ParseBuffer::integer()
       fail(__FILE__, __LINE__,"Expected a digit, got eof ");
    }
 
-   int signum = 1;
+   bool negative = false;
    if (*mPosition == '-')
    {
-      signum = -1;
+      negative = true;
       ++mPosition;
       assertNotEof();
    }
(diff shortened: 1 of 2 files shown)


rutil/ParseBuffer.cxx
```diff
@@ -671,25 +671,31 @@ ParseBuffer::integer()
...
+
+   // The absolute value limit depending on the detected sign
+   const unsigned int absoluteLimit = negative ? -(unsigned int)INT_MIN : INT_MAX;
+   // maximum value for all full range expect last digit
+   const unsigned int border = absoluteLimit / 10;
+   // value the last digit must not exceed
+   const unsigned int digitLimit = absoluteLimit % 10;
+
+   unsigned int num = 0;
+   while (!eof() && isdigit(*mPosition)) {
```
sgodin: Should clean up bracketing style here, then merging this pull request is good.   Thanks!

PR #113, opened 2018-05-12
rutil: Fix XMLCursor memory leak when ctor throws

gjasny:


sgodin:
Looks good to me.  Thanks Gregor!

PR #114, opened 2018-05-17
STUN: use RFC5389-fixed-magic-cookie to be able to decode XOR-MAPPED-…

HolgerJahnel:
…ADDRESS

Since RFC5389 stun server answer contains a XOR-MAPPED-ADDRESS version of MAPPED ADDRESS. To create the XOR-MAPPED-ADDRESS the 4 first bytes of transaction id field are used as 'magic cookie'.
RFC5389 chapter 6 says 'The magic cookie field MUST contain the fixed value 0x2112A442 in network byte order.'
Out there are stun servers, that don't use the random 4 bytes, delivered by the client, to XOR the MAPPED ADDRESS. They always use the fixed value from RFC5389, ignoring the first random 4 bytes in transaction id field.
In decoding algorithm resiprocate stun implementation uses the randomly generated 4 bytes, before this fix.
Now the fixed value 0x2112A442 for magic cookie is used, while transaction id is generated. So decoding is successful always.

### DIFF
--- a/rutil/stun/Stun.cxx
+++ b/rutil/stun/Stun.cxx
@@ -1918,6 +1918,7 @@ stunFindLocalInterfaces(UInt32* addresses,int maxRet)
 #endif
 }
 
+
 void
 stunBuildReqSimple( StunMessage* msg,
                     const StunAtrString& username,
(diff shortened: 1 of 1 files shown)


sgodin:
Thanks for this, and sorry for taking so long to respond.  This pull request has been obsoleted by a commit that was made in early 2021, and is no longer necessary.  This change:
https://github.com/resiprocate/resiprocate/commit/f1786ca9fde22f3b58a2a619744f626f40fd8bd9
accomplishes what this pull request was intending to do, as well adds support for parsing RFC5389 attributes.  

RFC5389
Section 6:  
   ...placing the magic cookie in this location allows
   a server to detect if the client will understand certain attributes
   that were added in this revised specification.

PR #115, opened 2018-06-18
Support LibreSSL

feinerer:
Newer versions of LibreSSL do not ship the `openssl/e_os2.h` header anymore. So exclude it if LibreSSL is detected.

### DIFF
--- a/resip/stack/ssl/DtlsTransport.cxx
+++ b/resip/stack/ssl/DtlsTransport.cxx
@@ -50,7 +50,10 @@
 
 #include "rutil/WinLeakCheck.hxx"
 
+#include <openssl/opensslv.h>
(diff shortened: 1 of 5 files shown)


sgodin:
I just tried commenting it out for Windows builds (using OpenSSL 1.0.2g),
and it built without any issues.  I wonder if older OpenSSL versions needed
that header for some reason.  The header appears to be setting up defines
for various different OS's.  If we remove it, might we lose support from
some OS's?

Scott

sgodin:
I am merging this pull request as is for now.  If we want to go back to see if e_os2.h can actually be removed entirely, then we can do that as a separate submission.  Thanks again Ingo!

PR #117, opened 2018-08-24
fixed linking problem with the latest gperf release (3.1)

szokovacs:


resip/stack/DateCategory.cxx
```diff
@@ -25,20 +25,34 @@ namespace resip
 struct days { const char *name; DayOfWeek type; };
 class DayOfWeekHash
 {
+#if  (GPERF_MAJOR_VER > 3 || (GPERF_MAJOR_VER == 3 && GPERF_MINOR_VER >= 1))
```
gjasny: As part of configuration, could you please add an `GPERF_SIZE_TYPE` and set it to `unsigned int` for older and `size_t` for newer gperf? Then use the `GPERF_SIZE_TYPE` here without ifdefs.

configure.ac
```diff
@@ -34,6 +34,17 @@ AC_PROG_RANLIB
 
 AM_MISSING_PROG([GPERF], [gperf])
 
+GPERFVER=`gperf -v|head -1 |cut -d " " -f3`
+GPERF_MAJOR_VER=`echo $GPERFVER | cut -f1 -d.`
+GPERF_MINOR_VER=`echo $GPERFVER | cut -f2 -d.`
+if test  $GPERF_MAJOR_VER -gt 3 -o $GPERF_MAJOR_VER -eq 3 -a $GPERF_MINOR_VER -le 1 ; then
```
gjasny: Shouldn't that be `$GPERF_MINOR_VER -ge 1` (>= instead of <=)?

resip/stack/HeaderHash.hxx
```diff
@@ -1,5 +1,6 @@
 #if !defined(RESIP_HEADERHASH_HXX)
 #define RESIP_HEADERHASH_HXX
+#include "config.h"
```
gjasny: You'll have to wrap it in
```c++
#if defined(HAVE_CONFIG_H)
#include "config.h"
#endif
```

```c++
#if defined(HAVE_CONFIG_H)
#include "config.h"
#else
#define GPERF_SIZE_TYPE unsigned int
#endif
```

configure.ac
```diff
@@ -34,6 +34,17 @@ AC_PROG_RANLIB
 
 AM_MISSING_PROG([GPERF], [gperf])
 
+GPERFVER=`gperf -v|head -1 |cut -d " " -f3`
```
gjasny: The version detection logic should take into account that gperf might be missing in which case the pre-generated (with 3.0.3) files from `resip/stack/gen` will be used.

resip/stack/HeaderHash.hxx
gjasny: Those header files get installed by autotools right now but `config.h` does not. Without a config file header file will differ from implementation. As far as I can see those `XyzHash.hxx` files are not referenced by any other files so we maybe could exclude them from install. @sgodin what's your take on this?

sgodin: HeaderHash.hxx is referenced by Headers.cxx
MethodHash.hxx is referenced by MethodTypes.cxx
ParameterHash.hxx is referenced by ParameterTypes.cxx

szokovacs: I think it would be better to install config.h as well, since:
szo@x:/usr/local/include$ rgrep config.h  |wc -l
37

szokovacs:
Anything else needed to merge?

gjasny:
Sorry, will take a look tomorrow.

resip/stack/DateCategory.cxx
```diff
@@ -1,5 +1,7 @@
 #if defined(HAVE_CONFIG_H)
 #include "config.h"
+#else
+#define GPERF_SIZE_TYPE unsigned int
```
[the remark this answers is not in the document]
szokovacs: yes, done.

configure.ac
```diff
@@ -34,6 +34,22 @@ AC_PROG_RANLIB
...
+  else
+    GPERF_SIZE_TYPE="unsigned int"
+  fi
+else
+  GPERF_SIZE_TYPE="unsigned int"
+fi
+
+AC_DEFINE_UNQUOTED(GPERF_SIZE_TYPE, $GPERF_SIZE_TYPE)
+AH_TEMPLATE([GPERF_SIZE_TYPE], [GPERF size type])
+
```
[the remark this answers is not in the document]
szokovacs: I could not find any.

szokovacs:
Looks like Travis has a problem with the libraries

configure.ac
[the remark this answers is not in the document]
dpocock: Could you please cut and paste the test code into a macro file in m4/gprefver.m4 perhaps?  This will help to keep the configure.ac file more concise.  You can find some tips here too:
https://autotools.io/autoconf/macros.html
https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.69/html_node/Writing-Autoconf-Macros.html#Writing-Autoconf-Macros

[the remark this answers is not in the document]
szokovacs: done
Travis still has trouble finding some apt packages :(

dpocock:
thanks for this contribution

Lastique:
This breaks linking with gperf 3.1. `GPERF_SIZE_TYPE` gets defined to `size_t`, which is `unsigned long` on x86-64. But the hash function implementations in `resip/stack/gen/*` use `unsigned int` for size arguments.

szokovacs:
> This breaks linking with gperf 3.1. `GPERF_SIZE_TYPE` gets defined to `size_t`, which is `unsigned long` on x86-64. But the hash function implementations in `resip/stack/gen/*` use `unsigned int` for size arguments.

Are you sure you regenerated the files in resip/stack/gen/ with 3.1? This whole patch was developed because for me, gperf 3.1 started to generate size_t.

Lastique:
No, I used the sources checked out from git and building with no modifications. The build system doesn't regenerate the files, if they are present. I think, that's the crux of the problem - autogenerated files need to be removed from git.

Lastique:
I've created https://github.com/resiprocate/resiprocate/pull/126 to fix this.

PR #121, opened 2018-10-25
Call-ID is recommended to use the form "localid@host" in RFC3261

Jerrylove:


sgodin:
Hi Jerry,

Thanks for the contribution!  However, I'm not sure it could be worded as "recommended" by the RFC, unless I missed something.  The only text in the RFC pertaining to this that I could find was:  
Implementations MAY use the form "localid@host".
Some sample call id's in the RFC use that form, other's do not. 

I'm not sure I see the value of merging this change, it will just make the messages a few bytes longer.

PR #122, opened 2019-01-22
Callback crash if destroy() is called whilst being in dns thread queue

bvandersson:


sgodin:
Hi Bjorn,

Thanks for the contribution.  Can you please describe in more detail how it is possible to trigger this crash to happen?  ie:  Did you reproduce this in repro or DUM, and what was going on at the time?  

Thanks,
Scott

bvandersson:
Hi Scott,

This i triggered by a bug on our side, we are sending an INVITE with UNKNOWN_TRANSPORT, and when we are canceling it a final dns lookup is called at reception of 487 in sendCurrentToWire().

My theory is that TransactionState mReliable is true in this case, and then in processClientInvite is sendCurrentToWire() called and later delete this;

Best Regards,

Björn

________________________________

PR #123, opened 2019-02-19
Fix std::string_view detection with newer Clang

feinerer:
Synced patch from
https://github.com/boostorg/asio/commit/43874d5497414c67655d901e48c939ef01337edb


sgodin:
Thanks Ingo!

PR #124, opened 2019-03-01
Two PR2

pengweichu:
1. Added mUserAgent member in the ContactInstanceRecord.
2. Add the parameter "removeOldFiles" for log which allows resip keep all log files.


sgodin:
Thanks for the contribution.  I will commit these separately after making some adjustments.

PR #125, opened 2019-03-22
rutil: Do not provide snprintf wrapper for VS2015 and newer

gjasny:
Defining snprintf breaks std::snprintf and should be avoided
wherever possible.

See: https://stackoverflow.com/a/8712996


sgodin:
Thanks Gregor!!

PR #127, opened 2019-04-16
Avoid using deprecated OpenSSL APIs

Lastique:
1. Use `ERR_remove_thread_state` to free thread-specific state on OpenSSL from 1.0.0 up to 1.1.0.
2. Use `SSL_COMP_free_compression_methods` on OpenSSL 1.0.2 to free memory allocated for compression methods.
3. Use `ASN1_STRING_get0_data` to access ASN.1 string contents. On OpenSSL older than 1.1.0, implement the function as a simple wrapper forwarding to `ASN1_STRING_data`.

This removes a bunch of compilation warnings about using deprecated OpenSSL APIs. In case of OpenSSL 1.0.2 this also fixes a memory leak of compression methods.

### DIFF
--- a/reTurn/AsyncTlsSocketBase.cxx
+++ b/reTurn/AsyncTlsSocketBase.cxx
@@ -7,6 +7,7 @@
 #include <boost/function.hpp>
 #include <boost/bind.hpp>
 
+#include <openssl/opensslv.h>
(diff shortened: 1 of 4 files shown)


sgodin:

PR #129, opened 2019-07-09
change the current startup directory as daemon process's working directory 

tangm421:
I think it is unreasonable to set `/` as the daemon process's working directory. while my config file at startup directory but the working directory at `/`, perhaps the directory `./` will be more convenient.

### DIFF
--- a/rutil/ServerProcess.cxx
+++ b/rutil/ServerProcess.cxx
@@ -267,7 +267,7 @@ ServerProcess::daemonize()
       // parent process done
       exit(0);
    }
-   if(chdir("/") < 0)
+   if(chdir("./") < 0)
    {
       ErrLog(<<"chdir() failed: "<<strerror(errno));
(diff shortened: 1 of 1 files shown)


dpocock:
Can you please tell us why you feel there is a problem using {{{/}}}, the root directory?  It is considered best practice for daemon processes to use {{{/}}}.
http://man7.org/linux/man-pages/man3/daemon.3.html

If you start it in some other directory then the other directory can't be deleted or unmounted.  It also ensures that each time you run the process, the environment is more consistent and any bugs you experience are less likely to be dependent on the directory you were in when starting it.

If this behavior causes problems for any use case then we can look at ways to resolve that, please consider posting a description of the problem to the mailing list:
http://list.resiprocate.org/mailman/listinfo/resiprocate-devel

tangm421:
i see，thank you

PR #131, opened 2019-11-01
ParseBuffer: Skip substring search in skipToChars when the length is …

rupprecht:
…longer than the amount of remaining characters.

This makes the loop terminate earlier, and also allows the compiler to safely optimize this to bcmp. See https://bugs.llvm.org/show_bug.cgi?id=43870 for more details on this optimization.

Tested by running the ParseBuffer test with `./configure CXXFLAGS=-fsanitize=address`


sgodin:
Thanks for the contribution!

PR #132, opened 2019-11-04
Fix off-by-one error in ParseBuffer::SkipToChars()

anderskristensen:
When the search string is present in the parse buffer, the buffer's position on return should be at
the beginning of where the search string was found. In the case where the search string was at the
very end of the buffer the new position would incorrectly be the end of the buffer.


sgodin:
Nice catch!

PR #137, opened 2019-12-18
Updated definition of header P-Access-Network-Info according to RFC 7…

BarfCliff:
…315, section 5.4.


sgodin:
Thank you for this contribution.  It looks good to me, but it would be ideal if we had a test case for the case.  Would you mind adding encode/decode tests for this change to resip/stack/test/testSipMessage.cxx?

BarfCliff:
Ok, I'll give it a try.

sgodin:
Great stuff - thanks!

BarfCliff:
Finally I could contribute with something, although a small part.

PR #139, opened 2020-02-06
Fix compilation error in resip::RecursiveMutex using musl on Alpine Linux

bwalden:
Found an issue when compiling rutil on Alpine Linux, which uses musl instead of glibc.

It appears that __linux__ is defined, but PTHREAD_MUTEX_RECURSIVE_NP is not. The change I've suggested appears to solve the problem because PTHREAD_MUTEX_RECURSIVE is defined. Happy to hear if you know of a better way though.


sgodin:
Looks good to me Ben, Thanks!

PR #141, opened 2020-03-16
Multiple crash scenarios involving the offer/answer state and an inco…

### LINKED ISSUE #140: Multiple crash scenarios involving the offer/answer state and an incoming request
1, Target refresh in the wrong invitesession state causes abort
(this was described in detail in the mailing list but was not concluded, see
https://list.resiprocate.org/archive/resiprocate-devel/msg09021.html
https://list.resiprocate.org/archive/resiprocate-devel/msg09022.html
https://list.resiprocate.org/archive/resiprocate-devel/msg09023.html
)
2, In case of SDP-less reINVITE, after receiving an offer, but failing to call provideAnswer() immediately (because for example, in case of a B2BUA, the other leg has to answer first), meanwhile receiving an UPDATE, InviteSession triggers an assert.

Both cases can be remedied by a well-placed 491.

szokovacs:
…ming request

fixes #140

### DIFF
--- a/resip/dum/Dialog.cxx
+++ b/resip/dum/Dialog.cxx
@@ -425,7 +425,17 @@ Dialog::dispatch(const SipMessage& msg)
          mNetworkAssociation.update(msg, keepAliveTime, false /* targetSupportsOutbound */); // target supports outbound is detected in registration responses only
       }
    }
-   
+
+   if(msg.isRequest() &&
+       (INVITE == msg.header(h_CSeq).method() || UPDATE == msg.header(h_CSeq).method()) &&
+       mInviteSession != 0 && mInviteSession->isUpdating())
+   {
+     SipMessage failure;
+     makeResponse(failure, msg, 491);
+     mDum.sendResponse(failure);
+     return;
+   }
+
    handleTargetRefresh(msg);
    if (msg.isRequest())
    {
--- a/resip/dum/InviteSession.cxx
+++ b/resip/dum/InviteSession.cxx
@@ -299,6 +299,29 @@ InviteSession::isConnected() const
 }
 
 bool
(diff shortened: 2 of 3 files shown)


dpocock:
@sgodin do you feel this is correct and that it has been inserted at the right place?

sgodin:

I see what this solution is trying to accomplish, but it is doing so outside of the InviteSession state machine.  There is already code in the state machine to accomplish the same thing.  ie:  See code in InviteSession::dispatchSentUpdate where a 491 response is sent.  It is confusing to have handling for this type of thing in two different places.  Handling in the state machine itself was the intended path for this type of thing.

@szokovacs Do you now exactly what scenarios were causing problems for you?  If so, we should be able to address this directly in the statemachine handling methods.

szokovacs:
Hi,
As I mentioned in
https://list.resiprocate.org/archive/resiprocate-devel/msg09023.html
the problem is with target refresh:
your patch in
https://list.resiprocate.org/archive/resiprocate-devel/msg09022.html
fixes the crash in dispatchOthers()
(all the other states mentioned in the proposed isUpdating() are handled correctly in their respective dispatch*() ), but it is too late, the contact is already overwritten.
It is actually mentioned in the comment in Dialog::handleTargetRefresh().

sgodin:
Thanks for the reminder.  :)  I think we need to go down one of the
following 2 paths:
1.  Figure out a solution for the target update problem, and handle in the
state machine as designed.     OR
2.  Use your solution, but remove all the state handling logic that will no
longer get used because of it.

I need some more time to ponder this.

Thanks,
Scott

szokovacs:
After discussing Scott, we agreed to go with a different solution.

PR #145, opened 2020-05-03
The DialogId can be const reference in the makeInviteSession to avoid…

portsip:
The DialogId can be const reference in the makeInviteSession to avoid the copy construction.


sgodin:
Thanks for this contribution.  :)

PR #146, opened 2020-05-11
In the Dialog::~Dialog(), delete the mClientSubscriptions and mServerSubscriptions incorrectly.

portsip:
The `mClientSubscriptions` and `mServerSubscriptions` both are list, we can't delete as that way.


sgodin:
Deleting the subscription will cause the item to be removed from the mClientSubscription list via the subscription destructor, so this code is working as designed.  ie:

ClientSubscription::~ClientSubscription()
{
   mDialog.mClientSubscriptions.remove(this);

Issue #147, opened 2020-05-12
Usage of std::auto_ptr prevents compilation of consuming projects when using C++17


bwalden:
As we know, C++17 removes std::auto_ptr. It looks like this is used in a few places throughout resiprocate.

I can see a couple of options to move forward:

1. Replace usages of std::auto_ptr with resip::SharedPtr.
2. Replace usages of std::auto_ptr with std::shared_ptr or std::unique_ptr as appropriate. This would make C++11 compiler support a requirement.

Both of the suggested options would mean a breaking change where std::auto_ptr is used publicly.

Any opinions or suggestions?

dpocock:
At this point, I feel that option (1) is better because it will ensure consistency through the project.  Later, we could then change all SharedPtr to one of the newer pointers.

gjasny:
Right now you can work-around by setting some defines (`_HAS_AUTO_PTR_ETC` for MSVC, `_LIBCPP_ENABLE_CXX17_REMOVED_AUTO_PTR` for libc++).

I'd vote against replacing the auto_ptrs with shared pointers for two reasons:
- You throw away life-cycle information. If you later convert to std::unique and std::shared pointers, you'll have to do some digging what the pointer type was before (and hope semantics did not change in between)
- You change life-cycle semantics and maybe introduce circular dependencies.

With the work-arounds to enable the auto_ptrs still available I'm thinking it's time to migrate the resiprocate code base to C++11 and start with replacing the auto_ptrs by std::unique_ptr and std::move.

sgodin:
Great points Gregor, I agree.  I think it's time we made the leap to C++11
as well.  Is there anyone that is aware offhand that this will break things
in their environment?

Scott

bwalden:
In light of this discussion and because I need this for a project I'm working on, I've implemented a change on my fork and submitted pull request #148 for review.

sgodin:
If we decide to make the leap to C++ 11 I can't see a reason to keep around
our resip::SharedPtr, if we are doing an ABI break with std::unique we may
as well break resip::SharedPtr and change to std::shared_ptr.  Thoughts?

Scott

gjasny:
Yes, I also think it should go away.

portsip:
Totally agree use the std::shared_ptr and std::unique_ptr.
[...]

PR #148, opened 2020-05-14
Replace std::auto_ptr with std::unique_ptr

bwalden:
As part of this, some classes which previously had a destructive copy constructor and assignment operator have had these explicitly deleted and replaced with a move constructor and assignment operator. Namely `resip::Helper::ContentsSecAttrs` and `dum::TargetCommand`.

The `dum::TargetCommand` class still has a destructive `clone()` method. This doesn't appear to cause a problem, the way it is used. But it may not be semantically correct.

I have not changed anything within the contrib directory or the p2p or tfm sub projects. These can be updated if required, but are not built by either the Travis script for Linux or the Visual Studio solution for Windows.


apps/ichat-gw/MediaRelay.cxx
```diff
@@ -330,10 +332,10 @@ MediaRelay::checkKeepalives(MediaRelayPort* relayPort)
       //}
 
       // Add message to buffer
-      std::auto_ptr<char> buffer(new char[sizeof(RtpHeader)]); 
+      std::unique_ptr<char> buffer(new char[sizeof(RtpHeader)]);
```
gjasny, line 335: do you have to use `std::unique_ptr<char[]>` here to invoke `delete []`?

resip/dum/TargetCommand.cxx
```diff
@@ -2,31 +2,27 @@
...
 void TargetCommand::executeCommand()
 {
-   mTarget.post(mMessage);
+   mTarget.post(std::move(mMessage));
 }
 
 Message* TargetCommand::clone() const
 {
-   return new TargetCommand(*this);
+   return new TargetCommand(mTarget, std::move(mMessage)); // !bw! this destructive clone is actually a move, which may not be right
```
gjasny, line 25: Is it actually used?

resip/stack/Helper.cxx
```diff
@@ -2206,12 +2182,12 @@ auto_ptr<SdpContents> Helper::getSdp(Contents* tree)
...
       {
          DebugLog(<< "Got sdp" << endl);
-         return auto_ptr<SdpContents>(static_cast<SdpContents*>(sdp->clone()));
+         return unique_ptr<SdpContents>(static_cast<SdpContents*>(sdp->clone()));
       }
    }
 
    //DebugLog(<< "No sdp" << endl);
-   return emptysdp;
+   return unique_ptr<SdpContents>();
```
gjasny, line 2190: `return nullptr`?

gjasny:
The autoconf should check for C++11 compatibility and enable it. Maybe @dpocock could help there?

resip/stack/Helper.cxx
bwalden: Agreed, nullptr would look tidier.

resip/dum/TargetCommand.cxx
bwalden: I think so. The `dum::EventDispatcher` and `post()` methods of `SipStack` both clone messages.

apps/ichat-gw/MediaRelay.cxx
bwalden: Yes, I believe so.

bwalden:
> The autoconf should check for C++11 compatibility and enable it. Maybe @dpocock could help there?

I assume we will have to add the following somewhere in configure.ac.
https://www.gnu.org/software/autoconf-archive/ax_cxx_compile_stdcxx.html

armatusmiles:
Hello. Any updates?

bwalden:
I think I've now made all of the suggested changes.

Possibly the only comment outstanding from issue #147, would be to also replace resip::SharedPtr with std::shared_ptr at the same time. This is certainly sensible and something I can do as part of this PR if it is wanted.

sgodin:
Thanks Ben!

It seems that no one is opposed to changing resip::SharedPtr to
std::shared_ptr, so I think it would be great to get this in the same PR,
since both changes will likely require app updates.

Scott

Scott

omnisip:
I've been wanting to get rid of resip::SharedPtr for years.  Please make
the change!  :-)

bwalden:
OK so resip::SharedPtr is now replaced with std::shared_ptr. I also took the liberty of replacing boost::shared_ptr with std::shared_ptr while I was in there.

Again, I've not touched the p2p or tfm sub projects. Does anyone have any thoughts on whether these should be updated?

portsip:
@bwalden 
Hi Bwalden, thanks for your works. Today I got the error when I try to compile your PR148 with Debian 10.4(64bit), GCC 8.3, please see this link: https://github.com/resiprocate/resiprocate/issues/147#issuecomment-677610287

BR

sgodin:
Hi Ben,

Thank you so much for all this work.  I was hoping to connect with you via email.  Can you please drop me a message at: s g o d i n @ s i p s p e c t r u m . c o m

Scott

bwalden:
Thank you all. It's nearly there now. I just want to finish off p2p and tfm. I decided to update these as well after finding they no longer compile following my changes.

portsip:
@sgodin @bwalden  Hi, a question, in the ClientPublish, since the req is a shared_ptr, why use std::move on it? If use std::move there then I suggest using std::unique_ptr. If use std::shared_ptr as the origin implementation (I mean before c+11), the std::move is not necessary.

sgodin:
> @sgodin @bwalden Hi, a question, in the ClientPublish, since the req is a shared_ptr, why use std::move on it? If use std::move there then I suggest using std::unique_ptr. If use std::shared_ptr as the origin implementation (I mean before c+11), the std::move does not make sense.

@portsip @bwalden  PortSIP - I agree with you.  Ben do you know why you used std::move here?  Did you do this elsewhere with shared_ptr's?  It could have unintended side effects.

bwalden:
You h> > @sgodin @bwalden Hi, a question, in the ClientPublish, since the req is a shared_ptr, why use std::move on it? If use std::move there then I suggest using std::unique_ptr. If use std::shared_ptr as the origin implementation (I mean before c+11), the std::move does not make sense.
> 
> @portsip @bwalden PortSIP - I agree with you. Ben do you know why you used std::move here? Did you do this elsewhere with shared_ptr's? It could have unintended side effects.

Yes this is using the modern C++ pass-by-value pattern. Any object which is passed by value to a constructor and is to be persisted as a member variable can be moved. This allows the compiler to optimize for us. If the object is an lvalue, the object is only copied once instead of twice. If the object is an rvalue, zero copies take place.

Some info on this in this example:
https://clang.llvm.org/extra/clang-tidy/checks/modernize-pass-by-value.html#:~:text=Pass%2Dby%2Dvalue%20in%20constructors,necessary%20to%20add%20this%20include.

The example uses std::string, but the principle for std::shared_ptr is the same.

```cpp
class MyFirstType
{
public:
    explicit MyFirstType(std::shared_ptr<Foo> foo)
        : m_foo(std::move(foo))
       // From this point on, foo is invalid and m_foo should be used
    {
    }

private:
    std::shared_ptr<Foo> m_foo;
};

void MyFirstFunc()
{
    auto foo1 = std::make_shared<Foo>();
    // foo1 is valid and has a ref count of 1

    MyFirstType mft1(foo1);
    // foo1 is still value and now has a ref count of 2

    auto foo2 = std::make_shared<Foo>();
    // foo2 is valid and has a ref count of 1

    MyFirstType mft2(std::move(foo2));
    // foo2 is invalid, mft2's m_foo member is has the only reference
}
```

Remember that performing a std::move on a std::shared_ptr is applying move semantics to the std::shared_ptr object itself, not the value pointed to by the std::shared_ptr.

sgodin:
Thank you Ben, for all your work with this pull request - it has now been merged with master.

dpocock:
@bwalden @sgodin there are some issues with the use of std::move here

I fixed one in recon and then I did a grep on reflow and found another one:

https://github.com/resiprocate/resiprocate/commit/7bb07cee4ac257dd7477adfa44836414da2bf285#diff-b3b2761215e678b972bd1b483df12aadccdff55fcc0db530bf7eb1bc60a4c2aa

In the case in recon: after the ```mProfile(std::move(profile))```, we can't use ```profile``` again, it is no longer valid, we can only use ```mProfile``` after that.  So I changed ```profile``` to ```mProfile``` in each case.

In the case in reflow: ```context``` is used twice in that branch of the ```if``` statement.  We can't use ```std::move``` twice or the second attempt is not valid.

On my platform the mProfile issue caused a segmentation fault

I'm concerned that there may be other examples of this lurking in the code but I didn't have time to do a full analysis.

dpocock:
Added this in Bugzilla as well so I can track it for the next release

https://www.resiprocate.org/bugzilla/show_bug.cgi?id=172

sgodin:
Yeah I was worried about the same thing.  In the few cases I found it was
obvious when running the unit tests as they caused exceptions when
misused.  I did a quick look around the code at the time and ran all our
unit tests and tfm tests.  I guess I missed a few cases.  We don't really
have any automated tests for recon, so that's probably how those went
unnoticed.  I've used the core stack and dum from master in a few projects
and haven't had issues, but I have done much with return, reflow or recon
since this work was completed.

Scott

dpocock:
Yes, I can see it was a lot of work and everybody who contributed to this deserves a lot of credit for helping modernize the project.

I just made an unrelated fix in recon and there was a non-deterministic build failure in rutil, only on gcc, the clang compile was fine:

https://travis-ci.org/github/resiprocate/resiprocate/builds/767161345

bwalden:
@dpocock @sgodin cast my eye over this again and I think I've found at least one other problem. Will fix and create a pull request.

PR #153, opened 2020-06-18
reject REFER request with Event: header if the corresponding handler …

szokovacs:
…is not registered

If a client sends a REFER request with a header like:
Event: notrefer

resiprocate would hit an assert and quit. This is undesirable, this change reject such requests with 491. If the application developer registered the event handle for "notrefer", nothing changes.


sgodin:
Hi Robert,  Someone else reported this issue as well, and I've been working off list on it a bit.  I'll email you the details.  I had a slightly different approach to fixing this.

szokovacs:
Commit b6d580dc6ed4aaab0b791e1ea1532253479178f1 has an alternative fix.

PR #154, opened 2020-06-19
Added Profile feature Handling491AsGeneralFailure

tvelencei:
The patch allows to switch off the resending mechanism activated by response 491 in order to allow the app layer to handle it. The feature is off by default. So if someone does not touch anything then DUM behaves as normal.

### DIFF
--- a/resip/dum/Profile.hxx
+++ b/resip/dum/Profile.hxx
@@ -236,6 +236,10 @@ class Profile
       virtual bool getExtraHeadersInReferNotifySipFragEnabled() const;
       virtual void unsetExtraHeadersInReferNotifySipFragEnabled();
 
+      virtual void setHandling491AsGeneralFailure(bool enabled);
+      virtual bool getHandling491AsGeneralFailure() const;
+      virtual void unsetHandling491AsGeneralFailure();
+
    private:
       bool mHasDefaultRegistrationExpires;
       UInt32 mDefaultRegistrationExpires;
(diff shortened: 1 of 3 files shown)


resip/dum/Profile.hxx
```diff
@@ -236,6 +236,10 @@ class Profile
       virtual bool getExtraHeadersInReferNotifySipFragEnabled() const;
       virtual void unsetExtraHeadersInReferNotifySipFragEnabled();
 
+      virtual void setHandling491AsGeneralFailure(bool enabled);
```
sgodin, line 239: To follow suit on the naming scheme already in place, and to reflect that this is for InviteSessions only, please rename this and the member variable to be:
setHandleInviteSession491AsGeneralFailureEnabled

sgodin:
Can you please explain what the use case is here, so I can understand it better.  Why would an application not want DUM to handle this automatically for them?

Thanks,
Scott

szokovacs:
> Can you please explain what the use case is here, so I can understand it better. Why would an 
> application not want DUM to handle this automatically for them?

We have a B2BUA that initiates a reinvite to both legs like this:
1, SDPless reINVITE to "C"
2, 200OK offer from "C"
3, forward the offer in reINVITE to "A"
4, 200OK answer from "A" (DUM handles the ACK)
5, forward the answer in ACK to "C"

If the 4 is not 200 but 491, the app must know about it and terminate the dialog toward "C" otherwise in some cases the remote client gets inpatient waiting for the ACK and terminates the whole call.

PR #157, opened 2020-08-07
Enable OpenSSL automatic curve selection mechanism

MilanKral:
SSL_{CTX_}set_ecdh_auto() has been removed from OpenSSL 1.1.0. and
ECDH is support is always enabled now. See
https://github.com/openssl/openssl/issues/1437#issuecomment-238828306

The code which calls SSL_CTX_ctrl(ctx, SSL_CTRL_SET_ECDH_AUTO
is failing with OpenSSL 1.1.0 and this causes set SSL_CTX_set_tmp_ecdh
with hard coded P-256 curve.

As a side-effect the OpenSSL automatic curve selection mechanism
is disabled and reSIProcate uses only the P-256 curve and doesn't use
newer curves, for example X-25519 with TLS connections.

### DIFF
--- a/resip/stack/ssl/Security.cxx
+++ b/resip/stack/ssl/Security.cxx
@@ -3154,6 +3154,8 @@ BaseSecurity::setDHParams(SSL_CTX* ctx)
       BIO_free(bio);
    }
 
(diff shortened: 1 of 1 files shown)


sgodin:
Thanks for the submission!

PR #163, opened 2020-11-18
autoconf: add ax_cxx_compile_stdcxx.m4

gjasny:
When we don't ship `ax_cxx_compile_stdcxx.m4` the `autoreconf --install && ./configure` might fail with:

```
./configure: line 11900: syntax error near unexpected token `11'
./configure: line 11900: `AX_CXX_COMPILE_STDCXX(11)'
```


sgodin:
Thanks Gregor!

PR #178, opened 2021-04-19
Fix an issue where the HEPSipMessageLoggingHandler constructor always throws

bwalden:
I've also run Clang-Tidy across the code base with all bugprone checks enabled and no more occurrences of bugprone-use-after-move have been detected for what it's worth.

There are quite a few other issues that have been highlighted though. I've attached the results. It might be worth going through to see if we need to worry about any of them.
[clang-tidy-results.txt](https://github.com/resiprocate/resiprocate/files/6334727/clang-tidy-results.txt)

I've got Jenkins running these checks. I'm not sure if Travis has a nice way to display the results. But it would be possible to run any worthwhile checks, write the results to the console, and fail the build if there are any problems found. If that's something that's of interest.

### DIFF
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,3 +1,4 @@
 Checks: >-
   -*,
   bugprone-*
+HeaderFilterRegex: .*
--- a/.clang-tidy
+++ b/.clang-tidy
@@ -1,0 +1,3 @@
+Checks: >-
+  -*,
+  bugprone-*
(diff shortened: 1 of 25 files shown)


sgodin:
This looks good to me - thanks for doing that extra work Ben!  @dpocock  it looks like build failure is with a unit test and only on one of the platforms.  I'm assuming it isn't caused by anything Ben did here.  If you are good with this, let's merge it in.

bwalden:
@sgodin the build failure looks to be that same issue @dpocock reported for testLogger in rutil for gcc.

sgodin:
Thanks Ben!

PR #183, opened 2021-06-17
resip/DUM: append stale flag to authenticate header

pbalash0v:
If `handleUserAuthInfo()` gets message with mode == UserAuthInfo::Stale than it is supposedly should lead to generation a `WWW-Authenticate` or `Proxy-Authenticate` which is akin to:
```
WWW-Authenticate: Digest nonce="1623965236:29487d401af226a46bc5e032992eb966",algorithm=MD5,realm="realm",qop="auth,auth-int",stale=true
```
(i.e. with a flag stale=true). But currently `stale=true` is not applied.
(further reference in [3.2.1 The WWW-Authenticate Response Header](https://datatracker.ietf.org/doc/html/rfc2617#section-3.2.1) in a `stale` paragraph)

AFAIU, the proposed way to change code keeps existing behavior intact, modifying handling only for UserAuthInfo::Stale case.

### DIFF
--- a/resip/dum/ServerAuthManager.cxx
+++ b/resip/dum/ServerAuthManager.cxx
@@ -205,7 +205,7 @@ ServerAuthManager::handleUserAuthInfo(UserAuthInfo* userAuth)
    {
       InfoLog (<< "Nonce expired for " << userAuth->getUser());
 
-      issueChallenge(requestWithAuth);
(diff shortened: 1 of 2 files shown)


sgodin:
Looks good to me.  Thanks for the contribution!

PR #188, opened 2021-08-10
Prevent the throw in constructor

xiexin0606:
Since the parameter agent is moved to the mHepAgent, the constructor always threw if determine the agent.


sgodin:
Thanks for the contribution!

sgodin (commented):
I'm going to have to revert this change, it doesn't build under VS2019:
1>C:\MyProjects\resip\rutil\hep\HepAgent.hxx(29,20): error C2898: 'void resip::HepAgent::sendToHOMER(const resip::TransportType,const resip::GenericIPAddress &,const resip::GenericIPAddress &,const resip::HepAgent::HEPEventType,const T &,const resip::Data &)': member function templates cannot be virtual

PR #190, opened 2021-08-17
Compatibles with Shared Address Space(RFC 6598)

xiexin0606:
Make the isPrivateAddress is compatibles with Shared Address Space(RFC 6598), for example, the  A-class IP 100.64.0.0－100.127.255.255.


sgodin:
Thanks for the contribution!

PR #201, opened 2021-12-13
Add Elliptic Curve Certificates as well as RSA

russhousley:
Add support for Elliptic Curve Certificates using the same names as the ones for RSA.  Also, update the RSA certificates to use SHA-256 for the CA certificate instead of SHA-1.


sgodin:
Thanks Russ!

PR #204, opened 2022-01-16
C++20 support on macOS

gjasny:
This PR makes resiprocate compile with C++20 on macOS. At should be backwards compatible down to C++11.

### DIFF
--- a/resip/stack/ParserContainer.hxx
+++ b/resip/stack/ParserContainer.hxx
@@ -177,10 +177,8 @@ class ParserContainer : public ParserContainerBase
             iterator operator++(int) {iterator it(mIt++,mRef); return it;}
             iterator operator--() {iterator it(--mIt,mRef); return it;}
             iterator operator--(int) {iterator it(mIt--,mRef); return it;}
-            bool operator!=(const iterator& rhs) { return mIt != rhs.mIt; }
-            bool operator==(const iterator& rhs) { return mIt == rhs.mIt; }
-            bool operator!=(const const_iterator& rhs) { return mIt != rhs.mIt; }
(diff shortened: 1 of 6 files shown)


resip/stack/ParserContainer.hxx
```diff
@@ -177,10 +177,8 @@ class ParserContainer : public ParserContainerBase
             iterator operator++(int) {iterator it(mIt++,mRef); return it;}
             iterator operator--() {iterator it(--mIt,mRef); return it;}
             iterator operator--(int) {iterator it(mIt--,mRef); return it;}
-            bool operator!=(const iterator& rhs) { return mIt != rhs.mIt; }
-            bool operator==(const iterator& rhs) { return mIt == rhs.mIt; }
-            bool operator!=(const const_iterator& rhs) { return mIt != rhs.mIt; }
-            bool operator==(const const_iterator& rhs) { return mIt == rhs.mIt; }
+            friend bool operator!=(const iterator& lhs, const iterator& rhs) { return lhs.mIt != rhs.mIt; }
+            friend bool operator==(const iterator& lhs, const iterator& rhs) { return lhs.mIt == rhs.mIt; }
```
gjasny, line 181: @miscco The (non-const) iterator and iterator should be comparable. But I struggle to provide a C++11 - 20 compatible solution without creating a mess. Do you have an idea for a minimal-invasive change?

miscco: I think the issue might be the rewriting operator.

I think this should work: https://godbolt.org/z/zrsTnW9YE

sgodin:
Thanks Gregor!  

I tested that this PR builds on VS2022.  I'll merge now, and if it causes issues on other compilers we can adjust.

Scott

PR #208, opened 2022-01-17
ci: add github action builds

gjasny:
Hello,

I noticed that currently reSIProcate does not run any CI builds. To add at least some basic checking I took the former Travis config and created a GitHub Actions build out of it. It has some shortcomings but it's better than nothing.

Follow-up tasks:

- Use Debian Bullseye instead of previous Buster release. Problem is that `libcajun-dev` has been dropped there
- reenable sipxtapi
- fix hanging / slow tests and re-enable testing
- Add mac and windows targets
- Add clang build (maybe macOS builds would cover that)
- speed up (compiler cache, pre-built docker image)
- Maybe use a dependency manager like vcpkg to provide dependencies on all operating systems

Thanks,
Gregor


sgodin:
Thanks Gregor!  Since Daniel did all the Travis CI work, I'll let him comment on this pull request.

gjasny:
@dpocock Looks like 63fb96b507a09f379fe2e9e98a3962768c29addf breaks the build. Could you please consider merging this PR to get working CI, again?

Issue #209, opened 2022-01-18
Repro's handling of Timer C


IvanRibakov:
As far as I can tell, current Repro behaviour when Timer C is fired is to cancel ALL client transactions:

```
// repro/ResponseContext.cxx
void
ResponseContext::processTimerC()
{
   if (!mRequestContext.mHaveSentFinalResponse)
   {
      InfoLog(<<"Canceling client transactions due to timer C.");
      cancelAllClientTransactions();
   }
}
```

```
INFO | 20220118-081343.009 | repro | REPRO:APP | 139697393129024 | RequestContext.cxx:895 | Updating timer C.
INFO | 20220118-081343.009 | repro | REPRO:APP | 139697393129024 | Proxy.cxx:569 | Posting timer C (8sec)
INFO | 20220118-081350.641 | repro | REPRO:APP | 139697393129024 | RequestContext.cxx:660 | RequestContext::process(ApplicationMessage) TimerCMessage(tid=-811889-1-0)
INFO | 20220118-081350.857 | repro | REPRO:APP | 139697393129024 | RequestContext.cxx:660 | RequestContext::process(ApplicationMessage) TimerCMessage(tid=-811889-1-0)
INFO | 20220118-081351.009 | repro | REPRO:APP | 139697393129024 | RequestContext.cxx:660 | RequestContext::process(ApplicationMessage) TimerCMessage(tid=-811889-1-0)
INFO | 20220118-081351.009 | repro | REPRO:APP | 139697393129024 | ResponseContext.cxx:981 | Canceling client transactions due to timer C.
INFO | 20220118-081351.009 | repro | REPRO:APP | 139697393129024 | ResponseContext.cxx:272 | Cancel ALL client transactions: 2 pending, 1 active.
```

Unless I'm misunderstanding something, that behaviour does not seem to be aligned with [RFC3261 which says following](https://datatracker.ietf.org/doc/html/rfc3261#section-16.8):

> 16.8 Processing Timer C
>
>
>   If timer C should fire, the proxy MUST either reset the timer with
>   any value it chooses, or terminate the client **transaction**.

Notice it says `terminate the client transaction` (singular).

Reason I'm bringing this up is because I'm trying to configure Repro to behave in a way that after a client transaction is terminated due to Timer C, next target in the delivery list is used to route the call to.

So, to summarise:
1. Is current Repro handling of Timer C valid?
2. Is it possible to continue routing call to other targets after one of them triggered Timer C?

[...]

PR #210, opened 2022-01-21
Running Repro's TimerC for each of the client TXs independently

IvanRibakov:
This PR addresses issue #209

### DIFF
--- a/repro/Proxy.cxx
+++ b/repro/Proxy.cxx
@@ -462,8 +462,24 @@ Proxy::thread()
                }
                else
                {
-                  InfoLog (<< "No matching request context...ignoring " << *app);
-                  delete app;
+                   RequestContextMap::iterator i = mClientRequestContexts.find(tid);
+                   if (i != mServerRequestContexts.end())
+                   {
+                       DebugLog(<< "Sending " << *app << " to " << *(i->second));
+                       try
+                       {
+                           i->second->process(std::unique_ptr<resip::ApplicationMessage>(app));
+                       }
+                       catch (resip::BaseException &e)
+                       {
+                           ErrLog(<< "Uncaught exception in process: " << e);
+                       }
+                   }
+                   else
+                   {
+                       InfoLog(<< "No matching request context...ignoring " << *app);
+                       delete app;
+                   }
                }
             }
             else if (term)
--- a/repro/Proxy.cxx
+++ b/repro/Proxy.cxx
@@ -463,7 +463,7 @@ Proxy::thread()
                else
                {
                    RequestContextMap::iterator i = mClientRequestContexts.find(tid);
-                   if (i != mServerRequestContexts.end())
+                   if (i != mClientRequestContexts.end())
                    {
                        DebugLog(<< "Sending " << *app << " to " << *(i->second));
                        try
--- a/repro/ResponseContext.cxx
+++ b/repro/ResponseContext.cxx
@@ -1013,7 +1013,8 @@ ResponseContext::processTimerC(const resip::Data &tid, int serial)
         InfoLog(<<"Canceling client transaction " << tid << " due to timer C.");
         cancelClientTransaction(tid);
 
-        if (serial == TimerCSerialInit) {
+        if (serial == TimerCSerialInit) 
+        {
             // TimerC fired BEFORE receiving provisional response
             // From RFC3261, 16.8 - 
             //      the proxy MUST behave as if the transaction received a 408(Request Timeout) response.
--- a/repro/ResponseContext.cxx
+++ b/repro/ResponseContext.cxx
@@ -984,7 +984,7 @@ void
 ResponseContext::updateTimerC(const resip::Data &tid)
 {
    InfoLog(<<"Updating timer C for client transaction " << tid);
-   int timerCSerial = 0;
+   int timerCSerial = TimerCSerialInit;
    auto i = mTimerCSerial.find(tid);
    if (i != mTimerCSerial.end())
    {
@@ -1001,28 +1001,35 @@ ResponseContext::updateTimerC(const resip::Data &tid)
 void
 ResponseContext::processTimerC(const resip::Data &tid, int serial)
 {
-   bool isLatestTimer = false;
-   auto i = mTimerCSerial.find(tid);
-   if (i != mTimerCSerial.end())
-   {
+    bool isLatestTimer = false;
+    auto i = mTimerCSerial.find(tid);
+    if (i != mTimerCSerial.end())
+    {
        isLatestTimer = (serial == i->second);
-   }
+    }
 
-   if (!mRequestContext.mHaveSentFinalResponse && isLatestTimer)
-   {
-      InfoLog(<<"Canceling client transaction " << tid << " due to timer C.");
-      cancelClientTransaction(tid);
+    if (!mRequestContext.mHaveSentFinalResponse && isLatestTimer)
+    {
+        InfoLog(<<"Canceling client transaction " << tid << " due to timer C.");
+        cancelClientTransaction(tid);
 
-      // After initiating client TX cancellation trigger next target processing  
-      auto i = mActiveTransactionMap.find(tid);
-      if (i != mActiveTransactionMap.end()) {
-          // Reconstruct original target request to be able to generate the response
-          auto targetRequest = buildTargetRequest(i->second);
-          std::unique_ptr<resip::SipMessage> response(Helper::makeResponse(*targetRequest, 408));
-          // Pretend like 408 response has been received to trigger next target processing
-          mRequestContext.process(std::move(response));
-      }
-   }
+        if (serial == TimerCSerialInit) {
+            // TimerC fired BEFORE receiving provisional response
+            // From RFC3261, 16.8 - 
+            //      the proxy MUST behave as if the transaction received a 408(Request Timeout) response.
+            // Otherwise, relying on the client to correctly handle CANCEL request and generate 487, which will
+            // also trigger next target processing
+            auto i = mActiveTransactionMap.find(tid);
+            if (i != mActiveTransactionMap.end())
+            {
+                // Reconstruct original target request to be able to generate the response
+                auto targetRequest = buildTargetRequest(i->second);
+                std::unique_ptr<resip::SipMessage> response(Helper::makeResponse(*targetRequest, 408));
+                // Pretend like 408 response has been received to trigger next target processing
+                mRequestContext.process(std::move(response));
+            }
+        }
+    }
 }
 
 void
--- a/repro/ResponseContext.cxx
+++ b/repro/ResponseContext.cxx
@@ -484,6 +484,101 @@ ResponseContext::isDuplicate(const repro::Target* target) const
    return false;
 }
 
+std::unique_ptr<resip::SipMessage>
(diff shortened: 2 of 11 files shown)


repro/ResponseContext.cxx
```diff
@@ -974,12 +981,47 @@ ResponseContext::processCancel(const SipMessage& request)
...
+
+      // After initiating client TX cancellation trigger next target processing  
+      auto i = mActiveTransactionMap.find(tid);
+      if (i != mActiveTransactionMap.end()) {
+          // Reconstruct original target request to be able to generate the response
+          auto targetRequest = buildTargetRequest(i->second);
+          std::unique_ptr<resip::SipMessage> response(Helper::makeResponse(*targetRequest, 408));
+          // Pretend like 408 response has been received to trigger next target processing
+          mRequestContext.process(std::move(response));
+      }
```
IvanRibakov, line 1024: Couldn't think of a better way to continue target processing than by pretending to receive a `408` response.

Are there alternative/better ways to do it?

repro/Proxy.cxx
```diff
@@ -462,8 +462,24 @@ Proxy::thread()
                }
```
IvanRibakov, line 462: https://github.com/resiprocate/resiprocate/issues/209#issuecomment-1016596117

> please also consider that there are some TFM test cases for TimerC that may need adjusting as well.

@sgodin - none of the existing TimerC tests seem to be failing, at least I got a PASS on the `tfm/repro/sanityTests.cxx` test suite.

I have tried to create a new TimerC test case with serial forking by using `testSequentialQValueInvite()` as a starting point, but I realised that it won't be as easy due to conflict between TimerC logic and `QValueCancelBetweenForkGroups`/`QValueMsBeforeCancel` configuration options. I couldn't find a way to tweak Repro configuration for a single test case so I abandoned the idea of the adding a new test case.

repro/ResponseContext.cxx
TheQue42: I appologize if I've missed something (but I am still trying to wrap my head around the code), but if timer C pops, that should trigger a CANCEL to the target, which in turn normally should trigger a 487 on the INVITE.
Sure, with bad/rfc2543-clients, you might just get 2xx on the CANCEL, (or worse, timeout), and in that case maybe a "fake" 408 is good.

From: 3261, 16.8

   If timer C should fire, the proxy MUST either reset the timer with
   any value it chooses, or terminate the client transaction.  If the
   client transaction has received a provisional response, the proxy
   MUST generate a CANCEL request matching that transaction.  If the
   client transaction has not received a provisional response, the proxy
   MUST behave as if the transaction received a 408 (Request Timeout)
   response.

IvanRibakov: > I appologize if I've missed something (but I am still trying to wrap my head around the stack/code as a whole)

No need to appologize, I'm in the same boat as you are - just getting started with resip/repro, evaluating it for potential use and probing for potential feature gaps.

Regarding your comment - I agree, it's a valid point. Ideally, there should be a check for the client TX state and `408` should only be generated if no provisional response has been received, otherwise we should expect client to do it's job and handle CANCEL as per RFC.

I've encountered implementation-level problem with the above approach - as far as I can tell, all TX state info is tracked inside `resip/stack`'s `TransactionController`, which is private to `SipStack` class and as such can not be accessed outside the `resip/stack`. That means that currently there is no easy way to check TX state from `repro`. I can think of some ways to work around that, I'm worried that doing so may affect some fundamental architecture elements.

As such, I would really appreciate some guidance from maintainers of this project regarding:
1. the current PR contents
2. analysis of the problem of accessing TX state from repro
3. potential exposure of `SipStack`'s transaction information (if not already possible)

repro/Proxy.cxx
sgodin: Right, I see what you mean.  QValueTargetHandler is created once and once the settings are passed in they are not changeable.  These are the options I can see:
1.  See if any tests rely on the QValueMsBeforeCancel setting being 5000 (from TestRepro.cxx line 49), if not then change it to what you need.
2. Add methods to QValueTargetHandler to allow the settings to be changed at runtime.
3. Add a way to restart the "proxy" under test, then in the test case change the value to what you need, restart the proxy, run the test, change back and restart the proxy again. 
4. Skip adding a unit test for this new behavior.

Options 1 or 2 seem best.

repro/Proxy.cxx
```diff
@@ -462,8 +462,24 @@ Proxy::thread()
...
                else
                {
-                  InfoLog (<< "No matching request context...ignoring " << *app);
-                  delete app;
+                   RequestContextMap::iterator i = mClientRequestContexts.find(tid);
+                   if (i != mServerRequestContexts.end())
```
sgodin, line 466: You are doing a find on mClientRequestContexts on line 465, but checking mServerRequestContexts.end() on 466.

repro/ResponseContext.cxx
sgodin: 1.  So far I like what I'm seeing.  Looks like  you handled the TCSerial logic nicely.  Great job for someone "just getting started". :)
2/3. Peeking into the TX state shouldn't be needed, we see all of the provisional responses in the ResponseContext code anyway.  In fact, we UpdateTimerC when a provisional arrives and bump the TCSerial.  So,  why not just check if the TCSerial value to see if a provisional has arrived or not (you already have access to it in this code)?

IvanRibakov: > we see all of the provisional responses in the ResponseContext code anyway

I didn't look much in this direction as I was reluctant to keep track of TX state in the `ResponseContext` to avoid complicating it further (aux variable init/cleanup + potentially duplicating some of the state tracking logic).

> So, why not just check if the TCSerial value to see if a provisional has arrived or not

Great idea! Will give it a go.

repro/Proxy.cxx
IvanRibakov: Well spotted, thanks!

repro/Proxy.cxx
IvanRibakov: 1. I've increased `QValueMsBeforeCancel` value from 5 seconds to 10 seconds and got 7 test failures (oddly enough, failures were only reflected in the `sanityTest.log` file, while testsuite summary in the console was suggesting all tests have passed). I could fix timeouts in the 7 affected tests and set TimerC value to be less than 10 seconds, but then it would change the test behaviour - client transactions would be cancelled sooner and due to TimerC expiry and  not because of the built-in QValueTargetHandler logic. This makes me think that a dynamic way to control Repro config is needed to maintain existing test logic separately from the new TimerC test logic.
2. Not sure how this can be done without exposing access to processor chains outside the `Proxy` class (which doesn't seem like a good idea on it's own).
3. Could probably be achieved but to ensure that one test failure doesn't cause a knock-on effect, it will probably be best to perform configuration reset to default values in the `ReproFixture::tearDown()`, which may add significant overhead to the overall test suite run time (due to Repro restart after each test case).

@sgodin not seeing any clear winners here (if you don't count option 4). Any thoughts on the above analysis? Am I missing something or are you willing to accept some of the above-mentioned trade-offs?

repro/ResponseContext.cxx
IvanRibakov: Done

repro/Proxy.cxx
sgodin: 1.  Seems odd you get failures that only show up in the logs and not in the summary.  Do those failures go away when you put the QValueMsBeforeCancel back to 5 seconds?  Is it possible to write a unit test for your new timerC logic that only dispatches one set of targets (ie: parallel forking), and not have QValueMsBeforeCancel get used.  I am assuming if there is no 2nd group of targets to dispatch then the it won't cancel, but I'm not sure offhand.
2. It should be pretty easy to just keep a pointer around to the QValueTargetHandler when it is added (TestRepro.cxx line 107/108), then call a method on it to change the settings when needed.
3. I don't like this suggestion anyway, let's forget it.  :)
4.  I'd like to ensure 1 or 2 aren't easy solutions before we go with this.

repro/ResponseContext.cxx
```diff
@@ -1001,28 +1001,35 @@ ResponseContext::updateTimerC(const resip::Data &tid)
...
-          // Pretend like 408 response has been received to trigger next target processing
-          mRequestContext.process(std::move(response));
-      }
-   }
+    if (!mRequestContext.mHaveSentFinalResponse && isLatestTimer)
+    {
+        InfoLog(<<"Canceling client transaction " << tid << " due to timer C.");
+        cancelClientTransaction(tid);
+
+        if (serial == TimerCSerialInit) {
```
sgodin, line 1016: Minor:  Please put all curly braces on their own line for all new code.

repro/ResponseContext.cxx
```diff
@@ -1001,28 +1001,35 @@ ResponseContext::updateTimerC(const resip::Data &tid)
...
+            // TimerC fired BEFORE receiving provisional response
+            // From RFC3261, 16.8 - 
+            //      the proxy MUST behave as if the transaction received a 408(Request Timeout) response.
```
sgodin, line 1019: This change looks good to me!  :)

tfm/repro/TestRepro.cxx
```diff
@@ -48,74 +48,14 @@ TfmProxyConfig::TfmProxyConfig(AbstractDb* db, const CommandLineParser& args)
    insertConfigValue("QValueMsBetweenForkGroups", "5000");
    insertConfigValue("QValueMsBeforeCancel", "5000");
 
+   insertConfigValue("TimerC", "30");
```
sgodin, line 51: I wonder if we should put this longer than the 32 second transaction timeout, just to make sure transaction timeouts have a chance to occur first.  ie:  Set to 40 instead?

tfm/repro/TestRepro.cxx
```diff
@@ -407,3 +413,15 @@ TestRepro::deleteTrustedHost(const resip::Data& host, resip::TransportType trans
        static_cast<const short&>(transport));
 }
 
+void 
+TestRepro::setQValueTargetHandlerCancelGroups(bool enabled)
```
sgodin, line 417: Looks great!  Nice work Ivan, and thanks for doing all this.

tfm/repro/TestRepro.cxx
```diff
@@ -322,6 +262,8 @@ TestRepro::TestRepro(const resip::Data& name,
 
 TestRepro::~TestRepro()
 {
+   mQValueTargetHandler = nullptr;
```
sgodin, line 265: We should be setting mQValueTargetHandler to nullptr in the constructor as well. :)

sgodin: Also, it's not really necessary to do it in the destructor, but I'm fine with leaving it there.

sgodin:
Great work Ivan!  Thanks so much for contributing to the project.  I will pull this code locally early next week and start using it on my test machine for a week to two before merging it with trunk.

tfm/repro/TestRepro.cxx
IvanRibakov: Good point, will test it out with the value of `40`.

IvanRibakov: Increased TimerC value to 40

PR #211, opened 2022-01-28
rutil: fix thread race in DnsUtil::getLocalHostName() static initialization

gjasny:
The previous code used only a single gate check before acquiring the lock,
which can lead to the case where multiple threads enter the
initialization block, with subsequent threads re-allocating the buffer of
the localHostName static member, leading to use-after-free memory reads.
This was detected by ASAN.

The fix is to rely on C++11 thread safe initialization guarantees for
local statics.

### DIFF
--- a/rutil/DnsUtil.cxx
+++ b/rutil/DnsUtil.cxx
@@ -113,73 +113,73 @@ DnsUtil::lookupARecords(const Data& host)
    }
 }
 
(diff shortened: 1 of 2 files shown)


gjasny:
Best reviewed w/o whitespace changes.

sgodin:
Thanks Gregor!

PR #212, opened 2022-01-31
Add support for dynamic timers B & C for each of the client transactions

IvanRibakov:
### Rationale

I'm evaluating Repro for it's suitability to be used as a foundation for ESRP ([Emergency Service Routing Proxy](https://www.nena.org/page/Glossary)). In the world of emergency service providers, it is not uncommon for providers, as well as governmental agencies overseeing those providers to express requirements for the upper bound of the call connection time. Those requirements are typically much stricter than the Timer B/Timer C values specified in the RFC 3261 and often vary based on the destination agency and a number of other call parameters.

As such, I'm looking for ability to specify custom TimerB and TimerC values for each of the Repro targets (serially forked). This PR proposes a solution to the problem.

### DIFF
--- a/repro/Proxy.cxx
+++ b/repro/Proxy.cxx
@@ -558,12 +558,16 @@ Proxy::addClientTransaction(const Data& transactionId, RequestContext* rc)
 }
 
 void
-Proxy::postTimerC(std::unique_ptr<TimerCMessage> tc)
+Proxy::postTimerC(std::unique_ptr<TimerCMessage> tc, int customDelayMs)
 {
-   if(mTimerC > 0)
+   // Convert ms to seconds
+   int delay = customDelayMs > 0 ? std::round(customDelayMs / 1000) : 0;
+   // Fall back to TimerC value from configuration if no custom delay is supplied
+   delay = delay > 0 ? delay : mTimerC;
+   if (delay > 0)
    {
-      InfoLog(<<"Posting timer C");
-      mStack.post(*tc,mTimerC,this);
+      InfoLog(<<"Posting timer C (" << delay << "sec)");
+      mStack.post(*tc, delay, this);
    }
 }
 
--- a/repro/ResponseContext.cxx
+++ b/repro/ResponseContext.cxx
@@ -492,7 +492,7 @@ ResponseContext::beginClientTransaction(repro::Target* target)
    resip_assert(target->status() == Target::Candidate);
 
    SipMessage& orig=mRequestContext.getOriginalRequest();
-   SipMessage request(orig);
+   SipMessage request(orig, target->mSipMessageOptions);
 
    // If the target has a ;lr parameter, then perform loose routing
    if(target->uri().exists(p_lr))
--- a/repro/ResponseContext.cxx
+++ b/repro/ResponseContext.cxx
@@ -584,7 +584,7 @@ ResponseContext::beginClientTransaction(repro::Target* target)
       mRequestContext.getOriginalRequest().method()==INVITE)
    {
       mRequestContext.mInitialTimerCSet=true;
-      mRequestContext.updateTimerC();
+      mRequestContext.updateTimerC(target->mSipMessageOptions.mTxOptions.mTimerC);
    }
    
    // the rest of 16.6 is implemented by the transaction layer of resip
@@ -1109,7 +1109,8 @@ ResponseContext::processResponse(SipMessage& response)
       case 1:
          if(mRequestContext.getOriginalRequest().method() == INVITE)
          {
-            mRequestContext.updateTimerC();
+            auto currentTargetPtr = i->second;
+            mRequestContext.updateTimerC(currentTargetPtr->mSipMessageOptions.mTxOptions.mTimerC);
          }
 
          if  (!mRequestContext.mHaveSentFinalResponse)
--- a/resip/stack/SipMessage.cxx
+++ b/resip/stack/SipMessage.cxx
@@ -65,6 +65,19 @@ SipMessage::SipMessage(const SipMessage& from)
    init(from);
 }
 
+SipMessage::SipMessage(const SipMessage &from, const SipMessageOptions &opts)
+    : mHeaders(StlPoolAllocator<HeaderFieldValueList *, PoolBase>(&mPool)),
+#ifndef __SUNPRO_CC
+      mUnknownHeaders(StlPoolAllocator<std::pair<Data, HeaderFieldValueList *>, PoolBase>(&mPool)),
+#else
+      mUnknownHeaders(),
+#endif
+      mCreatedTime(Timer::getTimeMicroSec())
+{
+    init(from);
+    mOpts = opts;
+}
+
 Message*
 SipMessage::clone() const
 {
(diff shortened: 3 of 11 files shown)


repro/Proxy.cxx
```diff
@@ -558,12 +558,16 @@ Proxy::addClientTransaction(const Data& transactionId, RequestContext* rc)
 }
 
 void
-Proxy::postTimerC(std::unique_ptr<TimerCMessage> tc)
+Proxy::postTimerC(std::unique_ptr<TimerCMessage> tc, int customDelayMs)
 {
-   if(mTimerC > 0)
+   // Convert ms to seconds
+   int delay = customDelayMs > 0 ? std::round(customDelayMs / 1000) : 0;
```
sgodin, line 564: FYI - You can use SipStack::postMS to post with a millisecond delay instead and not lose the precision on new method parameter units (if it must be MS for some reason).

resip/stack/SipMessage.hxx
```diff
@@ -737,6 +751,8 @@ class SipMessage : public TransactionMessage
 
       std::vector<MessageDecorator*> mOutboundDecorators;
 
+      SipMessageOptions mOpts;
```
sgodin, line 754: The SipMessage class is one we need to be really careful about to not add unnecessary bloat to.  It largely affects the memory use of any SIP server (as there can be 1000's or them allocated). If SipMessageOptions grows to be larger, and it is not used by most people, we'd probably be better off implementing it as dynamically allocated data, so that "extra" space only get's used if needed.  That being said, it only contains 2 int's now, which is the same size as a 64bit pointer, ...but I'm still thinking ahead to when it might get expanded for more uses.

resip/stack/TransactionMessage.hxx
```diff
@@ -17,7 +18,10 @@ class TransactionMessage : public Message
 
       // indicates this message is associated with a Client Transaction for the
       // purpose of determining which TransactionMap to use
-      virtual bool isClientTransaction() const = 0; 
+      virtual bool isClientTransaction() const = 0;
+
+      // Allows to customise TimerB value independently for each of the client TXs.
+      virtual int getTimerB() const { return Timer::TB; }
```
sgodin, line 24: seems strange you have SipMessageOptions setting for timerB, yet you don't return it here.  I'm assuming this should be returning SipMessageOptions timerB value if > 0, or Timer::TB otherwise?

sgodin: oops thought this was still the SipMessage class, disregard.

resip/stack/SipMessage.hxx
IvanRibakov: I see.

> but I'm still thinking ahead to when it might get expanded for more uses.

Would you expect both `SipMessageOptions` and `TxOptions` to be stored in a dynamic memory?

sgodin: I was thinking SipMessageOptions as a whole, hadn't really considered more fine grained....

repro/Proxy.cxx
IvanRibakov: The problem here is that TimerB is configured in milliseconds (in `Timer.cxx`) while TimerC is configured in seconds (in `repro.config`). I see 2 options here:
1. timer values are configured in different units to avoid additional conversion code. This could lead to some user/configuration errors in case of `TxOptions` where 2 timer values appear next to each other.
2. a common unit is chosen and extra conversion is performed

I've opted for the option 2 at the moment, but happy to switch to option 1 if you think the code simplicity outweighs the risk of misconfiguration.

sgodin: Actually I'd like to see the units in the variable names regardless, I think this really helps developers from needing to dig around to see how they are used.

repro/ResponseContext.cxx
```diff
@@ -584,7 +584,8 @@ ResponseContext::beginClientTransaction(repro::Target* target)
       mRequestContext.getOriginalRequest().method()==INVITE)
    {
       mRequestContext.mInitialTimerCSet=true;
-      mRequestContext.updateTimerC(target->mSipMessageOptions.mTxOptions.mTimerC);
+      int delay = target->mSipMessageOptions ? target->mSipMessageOptions->mTxOptions.mTimerC : 0;
```
sgodin, line 587: rename to delay to customDelayMs so that a value of 0 makes more sense when looking at this code :)

repro/ResponseContext.cxx
```diff
@@ -1110,7 +1111,8 @@ ResponseContext::processResponse(SipMessage& response)
          if(mRequestContext.getOriginalRequest().method() == INVITE)
          {
             auto currentTargetPtr = i->second;
-            mRequestContext.updateTimerC(currentTargetPtr->mSipMessageOptions.mTxOptions.mTimerC);
+            int delay = currentTargetPtr->mSipMessageOptions ? currentTargetPtr->mSipMessageOptions->mTxOptions.mTimerC : 0;
```
sgodin, line 1114: rename to delay to customDelayMs so that a value of 0 makes more sense when looking at this code :)

resip/stack/SipMessage.cxx
```diff
@@ -1820,7 +1826,7 @@ SipMessage::copyOutboundDecoratorsToStackFailureAck(SipMessage& ack)
 int 
 SipMessage::getTimerB() const
```
sgodin, line 1827: getTimerBMs?

resip/stack/SipMessage.cxx
```diff
@@ -75,7 +75,9 @@ SipMessage::SipMessage(const SipMessage &from, const SipMessageOptions &opts)
       mCreatedTime(Timer::getTimeMicroSec())
 {
     init(from);
-    mOpts = opts;
+    if (opts) {
```
sgodin, line 78: Bracket on separate line.  :)

Also trying to wrap my head around why we are copying the unique_ptr.  Using a unique_ptr gives the user the impression you are handing of the data to be managed by the class, so why copy here?  If the copy is necessary, we should think about changing the constructor signature to perhaps just take in a raw pointer.

IvanRibakov: Apologies for going quite for some time, other work stuff came up.

> Using a unique_ptr gives the user the impression you are handing of the data to be managed by the class

Not unless unique_ptr is passed by const reference (assuming no foul play by receiver of the const ref, like casting const away). 

> If the copy is necessary, we should think about changing the constructor signature to perhaps just take in a raw pointer.

I wanted to avoid using raw pointer to prevent it from being accidentally stored(copied) and outliving the unique_ptr that manages it.

When deciding on how best to pass `SipMessageOptions` to the constructor I used this [SO answer](https://stackoverflow.com/a/8114913/258825) for inspiration and reference. Let me know if reasoning above makes sense to you or if you'd like to switch to raw pointer approach in the end.

IvanRibakov: > If a copy is needed, then the caller can copy the unique_ptr before passing it in. Thoughts?

Done

----

> Do you intend to submit a new processor or some new functionality to repro to make use of these changes publicly?

I do not have such plans at the moment.

PR #217, opened 2022-02-25
Construct the document key without the port of URI.

pengweichu:


sgodin:
Could you please provide a reason for this change?  Is this to comply with RFC's or is there some other purpose?  Do you have some entities publishing with a port and subscribers trying to subscribe without a port, or vice-versa?  I'm thinking the URI matching we do here, should be the same as with Registration Uri matching.

One other point, please avoid placing 2 different changes in the same pull request.  It may hinder acceptance of the PR, if one change is accepted but another is not.  That being said - thank-you for the Network Association fix, it looks good to me.

sgodin:
FYI - I committed the Network Association fix separately from this.  

Will wait for your feedback on the publish/subscribe port issue before continuing with the review on that change.

Thanks!
Scott

sgodin:
Closing at request of submitter.

Issue #225, opened 2022-03-29
Helper::makeRawResponse generates invalid messages


bogdanvasilescu-opentext:
Helper::makeRawResponse does not append the CR LF sequence at the end of the start line of the message.

PR #226, opened 2022-03-29
Update Helper.cxx

bogdanvasilescu-opentext:
Added missing CR LF sequence at the end of the start line.

### DIFF
--- a/resip/stack/Helper.cxx
+++ b/resip/stack/Helper.cxx
@@ -498,7 +498,7 @@ Helper::makeRawResponse(Data& raw,
       encodeStream << "SIP/2.0 " << responseCode << " ";
       Data reason;
       getResponseCodeReason(responseCode, reason);
-      encodeStream << reason;
+      encodeStream << reason << Symbols::CRLF;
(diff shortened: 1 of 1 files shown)


bogdanvasilescu-opentext:
This change addresses issue #225

sgodin:
Wow!  Nice find.  I'm having trouble grasping why this hasn't been noticed before.  It seems that many 100 responses and some 503 responses that the stack generates uses this code.  May I ask how you noticed this?

bogdanvasilescu-opentext:
Hello @sgodin,

I was trying to see how the stack would handle a burst of new sessions (by firing 100 new invite sessions, immediately one after the other :D) using the repro app as proxy. 
In the sip messages log it looked like some of these invites were not receiving a response. After checking the main log I discovered these 503 responses where the stack was complaining about missing VIA header (although it was there). Eventually I realized that there was a problem with the message itself and not with the way it was logged.

sgodin:
Thanks for the explanation.  Very interesting.  :)

PR #229, opened 2022-04-07
reTurn: fix the async client to choose the DNS result corresponding to local endpoint protocol type

renaud-sc:
Currently, the fist result of the DNS query is used. I face a situation where the DNS query reported an IPv6 address first while my local endpoint (bond address) is IPv4 so that could not work.

This proposal select the first result where the protocol match the one of the local endpoint.

I also changed asio::ip::udp::resolver::iterator into asio::ip::udp::resolver::results_type as it is deprecated.

### DIFF
--- a/reTurn/AsyncUdpSocketBase.cxx
+++ b/reTurn/AsyncUdpSocketBase.cxx
@@ -79,10 +79,10 @@ AsyncUdpSocketBase::handleUdpResolve(const asio::error_code& ec,
          const asio::ip::udp::resolver::results_type::endpoint_type &ep = it->endpoint();
          if (ep.protocol() == mSocket.local_endpoint().protocol())
          {
-      mConnected = true;
+            mConnected = true;
             mConnectedAddress = ep.address();
             mConnectedPort = ep.port();
-      onConnectSuccess();
+            onConnectSuccess();
             return;
          }
       }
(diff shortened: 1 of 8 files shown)


reTurn/AsyncUdpSocketBase.cxx
```diff
@@ -67,17 +67,27 @@ AsyncUdpSocketBase::connect(const std::string& address, unsigned short port)
...
-      // Nothing to do for UDP except store the connected address/port
+      // Find the first endpoint in the list matching the local endpoint protocol type
+      asio::ip::udp::resolver::results_type::const_iterator       it  = results.begin();
+      const asio::ip::udp::resolver::results_type::const_iterator end = results.end();
+      for (; it != end; ++it)
+      {
+         const asio::ip::udp::resolver::results_type::endpoint_type &ep = it->endpoint();
+         if (ep.protocol() == mSocket.local_endpoint().protocol())
+         {
       mConnected = true;
```
sgodin, line 82: Some things look misaligned here.  It's hard to tell if it's just a github display issue.  Please make sure you are using spaces and not TABs for all resip changes, as this is the current coding convention.  see https://github.com/resiprocate/resiprocate/wiki/Code-Conventions

sgodin:
Looks great - thanks again Renauld!

PR #230, opened 2022-04-07
reTurn: allow to personalise the software attribut on TurnAsyncSocket

renaud-sc:


reTurn/client/TurnAsyncSocket.hxx
```diff
@@ -144,6 +148,8 @@ private:
    bool mCloseAfterDestroyAllocationFinishes;
    resip::Mutex mMutex;
 
+   std::string mSoftware;
```
sgodin, line 151: Let's use resip::Data as the storage type, to be consistent with other strings in this class (ie. mUsername, etc.)

reTurn/client/TurnAsyncSocket.hxx
```diff
@@ -66,6 +66,10 @@ public:
    // Sets the local HmacKey, used to check the integrity of incoming STUN messages
    void setLocalPassword(const char* password);
 
+   // Software attribute
+   void setSoftware(const std::string &software);
+   const char *getSoftware() const;
```
sgodin, line 71: the standard in this class is to place the pointer designation next to the type.  Can you please change this to be "const char* <method>" as opposed to "const char *<method>"?

reTurn/client/TurnAsyncSocket.cxx
```diff
@@ -1114,6 +1115,33 @@ TurnAsyncSocket::sendToRemotePeer(RemotePeer& remotePeer, const std::shared_ptr<
...
+
+      if (padded > unpadded)
+         mSoftware.append(padded - unpadded, ' ');
+   }
+}
+
+const char *
+TurnAsyncSocket::getSoftware() const
+{
+   if (mSoftware.empty())
```
sgodin, line 1139: This if seems a little odd, since you initialize mSoftware to SOFTWARE_STRING in the constructor.  Perhaps this is just protection incase someone set's to an empty string?  I think the logic should be different if someone set's it to an empty string though - I think we should not even add the software attribute in that case.  It is an optional attribute.

reTurn/client/TurnAsyncSocket.cxx
```diff
@@ -1114,6 +1115,33 @@ TurnAsyncSocket::sendToRemotePeer(RemotePeer& remotePeer, const std::shared_ptr<
    }
 }
 
+void
+TurnAsyncSocket::setSoftware(const std::string &software)
+{
+   mSoftware = software;
+
+   const size_t unpadded = mSoftware.size();
```
sgodin, line 1123: lets use the variable names unpaddedSize and paddedSize for more clarity

reTurn/client/TurnAsyncSocket.cxx
```diff
@@ -1114,6 +1115,33 @@ TurnAsyncSocket::sendToRemotePeer(RemotePeer& remotePeer, const std::shared_ptr<
...
+
+   if (unpadded > 0)
+   {
+      // Pad size to a multiple of 4, to help compatibility with older clients
+      const size_t remainder = unpadded % 4,
```
sgodin, line 1128: glad you thought to pad this :)

reTurn/client/TurnAsyncSocket.cxx
```diff
@@ -828,7 +829,7 @@ TurnAsyncSocket::handleBindRequest(StunMessage& stunMessage)
    StunMessage::setStunAtrAddressFromTuple(response->mXorMappedAddress, stunMessage.mRemoteTuple);
 
    // Add Software Attribute
-   response->setSoftware(SOFTWARE_STRING);
+   response->setSoftware(getSoftware());
```
sgodin, line 832: Further to comment below about not adding software header if set to empty, these set lines could be changed to the following and we can remove getSoftware():

if(!mSoftware.empty())
{
    response->setSoftware(mSoftware.c_str());
}

reTurn/client/TurnAsyncSocket.cxx
```diff
@@ -1114,6 +1115,33 @@ TurnAsyncSocket::sendToRemotePeer(RemotePeer& remotePeer, const std::shared_ptr<
    }
 }
 
+void
+TurnAsyncSocket::setSoftware(const std::string &software)
```
sgodin, line 1119: This API is not thread safe.  I would recommend that you follow the pattern established by the methods setLocalPassword and doSetLocalPassword.  This shows how to dispatch the setting of mSoftware to the thread where the string is used, and ensures you can change it at anytime in a thread safe manner.

sgodin:
TurnAsyncSocket and TurnSocket were meant to have the same API's.  One being synchronous and the other asynchronous.  Ideally this same change would be made in TurnSocket as well.  Are you able to include that in this PR?

renaud-sc:
I have modified the synchronous socket as well. I haven't been able to test with TLS but UDP and TCP seem fine.

sgodin:
Looks great - thanks for this contribution Renaud!

PR #234, opened 2022-04-14
reTurn client: revert DNS result filtering and pass local endpoint protocol to query object

renaud-sc:
Revert previous changes (except for some minor clean-up) and request the right protocol when initiating the DNS resolve query.

### DIFF
--- a/reTurn/AsyncTcpSocketBase.cxx
+++ b/reTurn/AsyncTcpSocketBase.cxx
@@ -48,15 +48,11 @@ AsyncTcpSocketBase::connect(const std::string& address, unsigned short port)
    // Start an asynchronous resolve to translate the address
    // into a list of endpoints.
    resip::Data service(port);
-#ifdef USE_IPV6
    asio::ip::tcp::resolver::query query(mSocket.local_endpoint().protocol(), address, service.c_str());
-#else
-   asio::ip::tcp::resolver::query query(asio::ip::tcp::v4(), address, service.c_str());   
-#endif
    mResolver.async_resolve(query,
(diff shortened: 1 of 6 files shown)


reTurn/AsyncTcpSocketBase.cxx
```diff
@@ -49,7 +49,7 @@ AsyncTcpSocketBase::connect(const std::string& address, unsigned short port)
    // into a list of endpoints.
    resip::Data service(port);
 #ifdef USE_IPV6
-   asio::ip::tcp::resolver::query query(address, service.c_str());   
+   asio::ip::tcp::resolver::query query(mSocket.local_endpoint().protocol(), address, service.c_str());
```
sgodin, line 52: is there a reason we can't just remove this #ifdef and always use mSocket.local_endpoint().protocol()?

reTurn/AsyncTlsSocketBase.cxx
```diff
@@ -67,7 +67,7 @@ AsyncTlsSocketBase::connect(const std::string& address, unsigned short port)
    // into a list of endpoints.
    resip::Data service(port);
 #ifdef USE_IPV6
-   asio::ip::tcp::resolver::query query(address, service.c_str());   
+   asio::ip::tcp::resolver::query query(mSocket.lowest_layer().local_endpoint().protocol(), address, service.c_str());
```
sgodin, line 70: Remove #ifdef?  This comment likely applies to a few other places.  I'll stop making it now.  :)

sgodin:
Thanks for the refactor Renaud - this is much cleaner.  :)

PR #237, opened 2022-05-03
Add top level namespace to things that conflicted with a seperate boo…

aerlfredith:


sgodin:
Thanks for the contribution, however, managing customizations to the asio headers is likely not something we want to start doing.

Issue #240, opened 2022-05-04
FileSystem::Directory::iterator does not calculate properly the mIsDirectory flag on Windows


bogdanvasilescu-opentext:
The Windows implementation of FileSystem::Directory::iterator uses incorrect attribute to figure out if current item is file or directory.
In addition, it does not initialize this flag in the constructor resulting in unpredictable behavior.

sgodin:
PR merged

PR #241, opened 2022-05-04
Update FileSystem.cxx

bogdanvasilescu-opentext:
FILE_ATTRIBUTE_DIRECTORY is the correct attribute to test in order to figure out if file find returned data is director or not.
In addition, mIsDirectory must be initialized in the constructor along with the other fields.


bogdanvasilescu-opentext:
This is a fix for #240

sgodin:
Thanks for this fix!

Issue #247, opened 2022-06-07
Mutex.h std::mutex cause crash on Windows 


hhmmdd:
When I do following change and recompile, my program works again. 
1.  change std::mutex to std::recursive_mutex in Mutex.h 
2. change std::condition_variable to std::condition_variable_any

jryden2:
@dpocock @sgodin I've successfully implemented this suggested change as well to resolve this issue.

gjasny:
When you say crash, do you mean a real or some (controlled) assertion terminating the process?

Could you please attach backtraces so that the underlying issue could be tracked down?

I cannot think of anything different than a deadlock that could be fixed by switching to recursive mutexes.

portsip:
> When you say crash, do you mean a real crash or some (controlled) assertion terminating the process?
> 
> Could you please attach backtraces so that the underlying issue could be tracked down?
> 
> I cannot think of anything different than a deadlock that could be fixed by switching to recursive mutexes.

As my remember, the crash is because initialize the global static std::mutex.

jryden2:
Hi Gregor (@gjasny),
  I can't speak for the OP, but my tests generate some kind of fatal exception accessing the new mutex.

` 	KernelBase.dll!00007ffd891c4fd9()	Unknown
 	vcruntime140d.dll!00007ffd09b9b660()	Unknown
 	msvcp140d.dll!00007ffc5ab84b93()	Unknown
 	msvcp140d.dll!00007ffc5ab84b0f()	Unknown
	App.dll!std::_Check_C_return(int _Res) Line 134	C++
 	App.dll!std::_Mutex_base::lock() Line 51	C++
 	App.dll!std::unique_lock<std::mutex>::unique_lock<std::mutex>(std::mutex & _Mtx) Line 134	C++
 	App.dll!resip::FdPollImplPoll::modPollItem(resip::FdPollItemFake * handle, unsigned short newMask) Line 592	C++
 	App.dll!resip::ConnectionManager::removeFromWritable(resip::Connection * conn) Line 174	C++
 	App.dll!resip::Connection::removeFrontOutstandingSend() Line 89	C++
 	App.dll!resip::Connection::performWrite() Line 289	C++
 	App.dll!resip::Connection::performWrites(unsigned int max) Line 299	C++
 	App.dll!resip::Connection::processPollEvent(unsigned short mask) Line 539	C++
[...]

PR #248, opened 2022-06-09
Mutex fixes

jryden2:
Resolves https://github.com/resiprocate/resiprocate/issues/247

### DIFF
--- a/rutil/Condition.hxx
+++ b/rutil/Condition.hxx
@@ -48,7 +48,7 @@ namespace resip
 
 // We removed our local implementation and now we use the
 // C++11 library
-typedef std::condition_variable Condition;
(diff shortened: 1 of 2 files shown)


portsip:
It makes no sense to use recursive_mutex instead of mutex - worth performance.

jryden2:
> It makes no sense to use recursive_mutex instead of mutex - worth performance.

Sure, that may be the case.  For now, this change resolves a complete failure.  Without it, the stack is unable to process any messages.  We may be able to isolate the recursive usage and only use the recursive_mutex where required, but this will require more investigation.

jryden2:
Good stuff.  Thanks Scott.

PR #250, opened 2022-06-27
Add support for building with OpenSSL 3.x

Lastique:
Removed the use of deprecated and removed APIs in OpenSSL 3.x. Also fix a few bugs in the affected code:

- Fixed incorrect delete operator being called for output buffers in `gruuUserPart` methods and use a fixed buffer in `SHA1Buffer`.
- Use SHA256 hash function in `BaseSecurity` identity functions, which matches the RSA signing hash function.
- Use constant strings when calling `Security::dumpAsn`.

### DIFF
--- a/resip/stack/Helper.cxx
+++ b/resip/stack/Helper.cxx
@@ -7,6 +7,8 @@
 #include <algorithm>
 #include <memory>
 #include <utility>
+#include <vector>
+#include <new>
 
 #include "resip/stack/Auth.hxx"
 #include "resip/stack/BasicNonceHelper.hxx"
@@ -1795,7 +1797,27 @@ Helper::validateMessage(const SipMessage& message,resip::Data* reason)
 }
 
 #if defined(USE_SSL) && !defined(OPENSSL_NO_BF)
+#include <openssl/opensslv.h>
 #include <openssl/blowfish.h>
+#if OPENSSL_VERSION_NUMBER >= 0x30000000L
+#include <openssl/evp.h>
+#include <openssl/params.h>
+#include <openssl/core_names.h>
+
+namespace {
+
+struct EVP_CIPHER_CTX_deleter
+{
+   typedef void result_type;
+
+   result_type operator() (EVP_CIPHER_CTX* p) const noexcept
+   {
+      EVP_CIPHER_CTX_free(p);
+   }
+};
+
+} // namespace
+#endif
 
 static const Data sep("[]");
 static const Data pad("\0\0\0\0\0\0\0", 7);
@@ -1818,9 +1840,6 @@ Helper::gruuUserPart(const Data& instanceId,
    ivec[6] = '\x7D';
    ivec[7] = '\x51';
 
-   BF_KEY fish;
-   BF_set_key(&fish, (int)key.size(), (const unsigned char*)key.data());
-
    const Data salt(resip::Random::getRandomHex(saltBytes));
 
    const Data token(salt + instanceId + sep + aor + '\0' +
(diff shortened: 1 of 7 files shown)


rutil/ssl/SHA1Stream.hxx
```diff
@@ -51,13 +41,8 @@ class SHA1Buffer : public std::streambuf
...
    private:
-      // !kh!
-      // used pointers to keep the same object layout.
-      // this adds overhead, two additional new/delete.
-      // could get rid of the overhead if, sizeof(SHA_CTX) and SHA_DIGEST_LENGTH are known and FIXED.
-      // could use pimpl to get rid of one new/delete pair.
-      std::unique_ptr<SHA_CTX> mContext;
-      std::vector<char> mBuf;
+      EVP_MD_CTX* mContext;
+      char mBuf[64];
```
gjasny, line 45: Could you include ` #include <openssl/evp.h>` and use:
```suggestion
      char mBuf[EVP_MAX_MD_SIZE];
```

rutil/ssl/SHA1Stream.cxx
```diff
@@ -18,27 +25,39 @@
...
+   {
+      EVP_MD_CTX_free(mContext);
+      mContext = NULL;
+      throw std::runtime_error("Failed to initialize SHA1 context");
+   }
 }
 
 SHA1Buffer::~SHA1Buffer()
 {
+   EVP_MD_CTX_free(mContext);
```
gjasny, line 48: For my code I created a deleter helper class:
```c++
struct OpenSslDeleter {
    void operator()(BIO* bio);
    void operator()(EVP_CIPHER_CTX* ctx);
    void operator()(EVP_ENCODE_CTX* ctx);
    void operator()(EVP_MD_CTX* ctx);
    void operator()(HMAC_CTX* ctx);
    void operator()(SSL* ssl);
    void operator()(SSL_CTX* ctx);
    void operator()(X509* x509);
};
```

With that I'm able to hold the OpenSSL contexts in RAII objects which enhances readability:
```c++
std::unique_ptr<EVP_CIPHER_CTX, OpenSslDeleter> m_ctx;
```

resip/stack/Helper.cxx
```diff
@@ -1855,33 +1917,72 @@ Helper::fromGruuUserPart(const Data& gruuUserPart,
...
                   BF_DECRYPT);
-   const Data pair(out.get(), decoded.size());
+#else
+   EVP_CIPHER_CTX* pCipherCtx = EVP_CIPHER_CTX_new();
+   if (!pCipherCtx)
+      throw std::bad_alloc();
+
+   std::size_t keylen = key.size();
+   OSSL_PARAM params[2];
+   params[0] = OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_KEYLEN, &keylen);
```
gjasny, line 1946:
```suggestion
   const OSSL_PARAM params[] = {
      OSSL_PARAM_construct_size_t(OSSL_CIPHER_PARAM_KEYLEN, &keylen),
      OSSL_PARAM_construct_end()
   };

File: rutil/ssl/SHA1Stream.hxx (hunk 1 above)
Comment (Lastique) [reply rutil/ssl/SHA1Stream.hxx:45]: This buffer is not used to store the digest, it is used for IO. E.g. if you output an integer into the stream, this is where it gets formatted.

The buffers for the digest are allocated in `SHA1Buffer::getHex` and `SHA1Buffer::getBin`, on the stack.

File: rutil/ssl/SHA1Stream.cxx (hunk 2 above)
Comment (Lastique) [reply rutil/ssl/SHA1Stream.cxx:48]: Yes, but that requires C++11. Is libresiprocate requiring C++11?

File: resip/stack/Helper.cxx (hunk 3 above)
Comment (Lastique) [reply resip/stack/Helper.cxx]: Done.

File: rutil/ssl/SHA1Stream.cxx (hunk 2 above)
Comment (sgodin) [reply rutil/ssl/SHA1Stream.cxx:48]: Yes, recent commits to the repo, now require C++ 11 for resip (ie: std::lock, std::mutex, etc).

Comment (Lastique) [reply rutil/ssl/SHA1Stream.cxx]: Ok, done.

**Comment by Lastique** [comment]
Ping? Any other comments?

**Comment by gjasny** [comment]
@Lastique Thank you for your work. It took a little bit longer to review than expected. I picked it piecewise onto master. The only thing left is merging #268 (if @sgodin) agrees and the `BaseSecurity` stuff. 

There I wonder if we could drop support for OpenSSL older than 1.0.2 to clean up the code. Personally I'd like to raise the required OpenSSL version to 1.1.1. @sgodin @dpocock What's your opinion on OpenSSL compatibility?

File: resip/stack/ssl/Security.cxx [hunk 4]
```diff
@@ -1913,19 +1938,41 @@ BaseSecurity::computeIdentity( const Data& signerDomain, const Data& in ) const
       return Data::Empty;
    }
 #endif
+#else // OPENSSL_VERSION_NUMBER < 0x30000000L
+   std::unique_ptr<EVP_PKEY_CTX, EVP_PKEY_CTX_deleter> pKeyCtx(EVP_PKEY_CTX_new(pKey, nullptr));
```
gjasny, line 1942: From a quick glance the `EVP_PKEY` API is available since OpenSSL 1.0.2. If we drop support for earlier OpenSSL releases we could exclusively use the new code.

Lastique:
I'm fine with raising the minimum OpenSSL version, but there are quite a few places that need to be cleaned up - some of them are not touched by this PR. So maybe the version requirement bump (and the associated cleanup) should be done separately.

sgodin:
> Personally I'd like to raise the required OpenSSL version to 1.1.1

I have no issues with that.  Thanks for reviewing all this work Gregor, and thanks for the contribution @Lastique.

gjasny:
With the merge of #280 resiprocate master should be OpenSSL 3 clean. One thing I have not merged is the addition of non-deprecated OpenSSL v3 API. I will do that in a future PR. But at the moment there is no urgent need to do so because we target OpenSSL 1.1.1 API which is still present in 3.0. And the non-deprecated API is not present in 1.1.1, yet.

Thank you for your help,
Gregor

Lastique:
You didn't merge the fix in `resip/stack/ssl/Security.cxx` - `BaseSecurity::computeIdentity` and `BaseSecurity::checkIdentity` still use `SHA1Stream` to compute the digest which is then used with a SHA256 signature.

gjasny:
You’re right. OpenSSL [man page for RSA_sign](https://www.openssl.org/docs/man1.1.1/man3/RSA_sign.html) states:

> type denotes the message digest algorithm that was used to generate m.

I will create a PR with your fix

PR #268, opened 2022-09-10
resip/stack: drop unused gruuUserPart functions

gjasny:
@sgodin Those functions look unused and lack any tests. Instead of porting them to OpenSSL 3 could we get rid of them completely.

(See #250)


sgodin:
I see a number of comments in the code the GRUU is a work in progress and full support is still a TODO.  Instead of deleting the code, I would leave it present in the code base and just wrap in openssl < 3.0.0, with a comment as to why it wasn't ported.  That way if we end supporting GRUU more in the future, the methods could be ported at that time.

gjasny:
The comment has been made in 2004 (a293bc38a6). Do you still want to keep the code?

sgodin:
The GRUU RFC went draft in 2009, I suspect (but haven't confirmed) that the code is still relevant.  If we delete it entirely it's not likely to be located again, if someone wants to implement it.

sgodin:
It's also possible that someone is using that API in their application.  I see no big motivation to out right delete it.

Lastique:
If the code is useful and potentially used downstream, I think it should be available with OpenSSL 3.0. You can use the work on porting done in https://github.com/resiprocate/resiprocate/pull/250.

Leaving the code non-functional just to keep it around doesn't sound like a useful approach to me.

PR #278, opened 2022-11-08
rewrite #ifdef/#elif to #if define(

ehasting:


gjasny:
Looking forward to C++23 `#elifdef`

PR #295, opened 2022-12-19
rutil: use C++11 hash map

gjasny:


sgodin:
Thanks Gregor, this looks good to me.  A nice bit of code cleanup too!  I'll leave as a PR a bit longer to let others review, then we can merge.

PR #300, opened 2023-01-24
VS solution configuration fixes

bwalden:
- Change the SSL-Release|x64 solution configuration to use the SSL-Release|x64 ares project configuration


sgodin:
I will merge this, but it should be noted that it shouldn't have been causing any issues.  ares doesn't use OpenSSL, I'm not 100% sure how the projects ended up with SSL configurations, but I suspect someone at sometime ended up propagating the solution configurations to the project configurations. In summary the SSL and non-SSL configurations for ares should have the exact same settings.   Since the configuration is there, we may as well be using it though.   Thanks Ben.

bwalden:
I agree. I only spotted this because I have a Conan recipe for reSIProcate that I was updating for Conan 2.0. Having ares use a different project configuration for that specific configuration/platform combo meant that ares didn't pick up Conan's new MSBuildToolchain.

PR #321, opened 2023-07-20
fix SDL compatibility

sherlock1982:
Visual Studio compiler has special SDL option to enforce certain additional checks. 
[Read about it here](https://learn.microsoft.com/en-us/cpp/build/reference/sdl-enable-additional-security-checks?view=msvc-170)

Resiprocate code almost complies with it excluding these two issues. 
First one is a mere warning suppress.

Second is more problematic if returned error is not expected pAdapterAddresses is undefined and app will crash


sgodin:

PR #323, opened 2023-08-08
delete msg after use

sherlock1982:
Note the line 229 where _msg is used after msg already deleted


sgodin:
Another great find - thanks again!

PR #324, opened 2023-08-08
fix header clear

sherlock1982:
Should be clear instead of empty?


sgodin:
Definitely - nice find, thanks!

PR #326, opened 2023-08-25
Fix build with VisualStudio 17.7.2 - Windows SDK 10.0.22621

renaud-sc:


sgodin:
Thanks Renaud!  Out of curiosity, do you remember what the compilation error was?

renaud-sc:
Hi Scott, here is a screenshot:

resip build failure

sgodin:
Interesting I don't get this error with version 17.6.2 and SDK version 10.0.22000.0, but I don't see any harm in adding the include, so I'll merge in.  Cheers!

renaud-sc:
Thanks Scott. 

I confirm it works fine on VS 17.7.3 as well (yes we get a new VS version every week nowdays). Probably more related to Windows SDK then VS anyway.

PR #336, opened 2023-10-19
rutil: include ares provided nameser defines

gjasny:
@sgodin Could you please check that with this change the code still compiles for you on Windows?


sgodin:
@gjasny Tested on Visual Studio 2022 and all looks good.  I included an extra change that get's rid of a bunch of compiler warnings that weren't related to your changes.  :)

PR #339, opened 2023-10-25
Feature/vcpkg

renaud-sc:
Add VCPKG support for CMake.

Could be handy for projects using VCPKG and causes no harm to the ones who don't.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,10 @@
 cmake_minimum_required(VERSION 3.5 FATAL_ERROR)
+
+option(USE_VCPKG "assume vcpkg presence and include it so find_package finds it." FALSE)
+if(USE_VCPKG)
+  include(${VCPKG_DIR})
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -1,4 +1,10 @@
 cmake_minimum_required(VERSION 3.6 FATAL_ERROR)
+
+option(USE_VCPKG "assume vcpkg presence and include it so find_package finds it." FALSE)
+if(USE_VCPKG)
+  include(${VCPKG_DIR})
```
gjasny, line 5: Is this a pattern the vcpkg devs recommend? I believe you could achieve the same thing with the CMake [CMAKE_PROJECT_TOP_LEVEL_INCLUDES](https://cmake.org/cmake/help/latest/variable/CMAKE_PROJECT_TOP_LEVEL_INCLUDES.html) feature.

renaud-sc: Oh ! We didn't know about that feature. Many thanks for the tips !
I will close this pull request.

PR #348, opened 2024-03-06
Push rejection for multiple values in a single-value header into PEDANTIC_STACK mode

bwalden:
Came across a version of ASBCE which produces requests containing multiple Content-Disposition values.

We're unable to get the ASBCE fixed, and while this behaviour is non-conformant; it doesn't really cause a problem. In the interest of interoperability, I've pushed the validation into PEDANTIC_STACK mode and ensured that only the first value seen is used. Added a test to confirm.


sgodin:
Thanks Ben!

PR #349, opened 2024-03-15
Improve library version handling

vak-oskh:
The main idea is to provide the ability to understand from third-party applications which a class method can be used depending on resiprocate library version.

I am upgrading the resiprocate library for my code from version 1.9.2 to master (possibly 1.13.0) and found the issue: `resip::SipMessage::getTransport()` vs `resip::SipMessage::getReceivedTransportTuple()`. I think there will be more )

The PR changes allows to include the header `<resip/version.h>` header and find the library version by the defined variable `RESIPROCATE_VERSION_NUMBER` or function `RESIPROCATE_VERSION_PREREQ(maj, min, patch)`.
For example,
```
const resip::Tuple &tuple = 
#ifdef RESIPROCATE_VERSION_PREREQ(1,13,0)
   *sipMsg.getReceivedTransportTuple();
#else // RESIPROCATE_VERSION_1_9_2
   sipMessage.getReceivedTransport()->getTuple();
#endif
```

Meanwhile, the resiprocate library staff can use the header `<version.h>` with the defined variable HAVE_VERSION_H for any cases where the version number is required.

P.S. The idea was borrowed from the OpenSSL library with the definition of OPENSSL_VERSION_NUMBER.


vak-oskh:
Dear maintainers.

I would like to ask you to take a look at my pull request.
I am using it during my upgrade and it is quite useful.

Thanks in advance.

vak-oskh:
Hello @sgodin.

I totally agree with your thought.
I would be happy to hear any feedback and provide appropriate fixes if they are required.

Thank you for your answer. 
I look forward to hearing any comments regarding this matter.

PR #360, opened 2024-06-22
rutil: add a public getter for the percent of maximum tolerance for C…

vak-oskh:
…ongestion Manager

Any third-party statistic will be able to read this data.


vak-oskh:
Just to make things clear.
The changes in the patch have been made for the following methods `getCongestionPercent` and `getCongestionPercentInternal`. The Git patch view shows it in a little bit different way for the header file.

sgodin:
Thanks for the contribution.

PR #361, opened 2024-07-19
rutil: fix compile error with MSVC

fnMrRice:
Fix compile error `C2121` on file `Log.cxx`. 
The `#ifdef` on `GenericLog` macro caused this issue. 
Moving those fields out would solved it.


sgodin:
Thanks for this contribution!

PR #362, opened 2024-07-19
resip/stack: allow to select transport if the multiple interfaces are configured for the same protocol

vak-oskh:
The issue has been found that a service based on resiprocate transport selection does not find a transport if multiple interfaces are configured for the same protocol. For example,
```
TransportSelector.cxx:222:Adding transport: [ V4 192.168.248.39:5065 UDP transportKey=1 ]
TransportSelector.cxx:222:Adding transport: [ V4 192.168.248.39:5075 UDP transportKey=2 ]
TransportSelector.cxx:222:Adding transport: [ V4 192.168.248.39:5065 TCP transportKey=3 ]
TransportSelector.cxx:222:Adding transport: [ V4 192.168.248.39:5075 TCP transportKey=4 ]
```

The SIP requests that are generated by the service, when the transport key is not detected and the Via header is expected to be added during transport selection could be rejected as no transports detected.
For example, the transport selection with the mentioned transport configuration and the following SIP request:
```
OPTIONS sip:192.168.248.40:5060;transport=udp SIP/2.0
Via: SIP/2.0/ ;branch=z9hG4bK-524287-1---d53a3635f7cd6c55;rport
Max-Forwards: 70
Route: <sip:192.168.248.40;transport=tcp;lr>
Contact: <sip:Test@192.168.248.39:5065>
To: <sip:192.168.248.40:5060;transport=udp>
From: <sip:192.168.248.39:5065>;tag=4d603416
Call-ID: 7tKr-9fy2O7H3VDV6VSF0w..
CSeq: 1 OPTIONS
Content-Length: 0
```
, the processing logs are:
```
TransportSelector.cxx:916:Looked up source for destination: [ V4 192.168.248.40:5060 UDP targetDomain=192.168.248.40 ] -> [ V4 192.168.248.40:0 UDP targetDomain=192.168.248.40 ] sent-by= sent-port=0 
TransportSelector.cxx:1556:findTransportBySource([ V4 192.168.248.40:0 UDP targetDomain=192.168.248.40 ]) 
TransportSelector.cxx:1570:should port be ignored: 1 
TransportSelector.cxx:1639:Exact interface / Specific port: [[ V4 192.168.248.39:5065 TCP transportKey=3 ] -> 0x2098990, [ V4 192.168.248.39:5075 TCP transportKey=4 ] -> 0x209aec0, [ V4 192.168.248.39:5065 UDP transportKey=1 ] -> 0x2076790, [ V4 192.168.248.39:5075 UDP transportKey=2 ] -> 0x2087820] 
TransportSelector.cxx:1640:Any interface / Specific port: [] 
TransportSelector.cxx:1641:Exact interface / Any port: [[ V4 192.168.248.39:5065 TCP transportKey=3 ] -> 0x209aec0, [ V4 192.168.248.39:5065 UDP transportKey=1 ] -> 0x2087820] 
TransportSelector.cxx:1642:Any interface / Any port: [] 
TransportSelector.cxx:1644:Can't find matching transport [ V4 192.168.248.40:0 UDP targetDomain=192.168.248.40 ] 
TransportSelector.cxx:1357:tid=d53a3635f7cd6c55 failed to find a transport to [ V4 192.168.248.40:5060 UDP targetDomain=192.168.248.40 ] 
```

The provided patch allows for TransportSelector to choose the first configured transport with the provided protocol.
The tests for TransportSelector have been restored:
* tests for `findTransportBySource()` were reworked and extended by cases for loopback and IPv6 addresses.
* tests for `findTransportByDest()` were added for general and mentioned cases.


resip/stack/test/CMakeLists.txt
vak-oskh, line 1: A question to the reviewer or maintainer. Shouldn't this test be run for the Windows platform as it has been done for testTransactionFSM?
I don't know these requirements as I've tested on Linux.

resip/stack/test/CMakeLists.txt
```diff
@@ -59,6 +59,7 @@ test(testTimer testTimer.cxx)
 if(NOT WIN32)
 manual_test(testTransactionFSM testTransactionFSM.cxx TestSupport.cxx)
 endif()
+test(testTransportSelector testTransportSelector.cxx)
```
vak-oskh, line 62: A question to the reviewer or maintainer. Shouldn't this test be run for the Windows platform as it has been done for testTransactionFSM?
I don't know these requirements as I've tested on Linux.

sgodin: I'm not sure the history, but testTransactionFSM does not build on Windows, so it is excluded from Windows builds - ideally this would be fixed.  The old testTransportSelector won't compile on Windows either, I suspect it won't compile on linux either, since it is trying to construct TransportSelector with the wrong number of arguments.  Once this change get's accepted, I'll see if the new testTransportSelector will compile/run on windows, and if not I can exclude it for windows until someone has cycles to address.

sgodin:
To understand the current transport selector process (without this PR), I created some pseudo code logic:

TransportSelector::transmit
  If Request
    findTransportByVia - check top via for matching source host and port that matches a transport
    If not found -> findTransportByDest
      If key set then select matching transport
      If target matches (AnyPortAnyInterfaceCompare) exactly one transport by type, ie: target is TCP it will match first TCP transport
    **If not found -> determineSourceInterface - takes the destination and queries the OS for the best interface to use to send to that destination**
      -> findTransportBySource - find a transport that matches the source interface found above
  If Response
    findTransportByDest
      If key set then select matching transport
      If target matches (AnyPortAnyInterfaceCompare) exactly one transport type

From what I can see, this PR will render the bolded logic above useless, since findTransportByDest will pretty much always return a transport (assuming one is added for each transport type).  If this is the case, this would be a big issue for transport selection, when multiple interfaces are used.  It will never query the OS to ask for the best interface to route to a particular target.

I'm guessing you are having issues because the OS is reporting (in determineSourceInterface) that the source interface to reach 192.168.248.40 is actually an IP address owned by this machine and thus source interface is 192.168.248.40, and you didn't actually add a transport on 192.168.248.40...
TransportSelector.cxx:1556:findTransportBySource([ V4 192.168.248.40:0 UDP targetDomain=192.168.248.40 ]) 
Is it true that 192.168.248.40 is an IP address on the machine you are running on?

vak-oskh:
Hello @sgodin,

I wanted to say thank you for the explanation.
I see that my first point of view for `TransportSelector::transmit()` explains the logic of `TransportSelector::determineSourceInterface()` in the wrong way.
In this case, I agree with you that my patch makes `determineSourceInterface()` and `findTransportBySource()` useless.
By the way, it means that `TransportSelector::findTransportByDest()` is a helper to find the transport with a simple configuration as it was designed to work only with a single transport per type but not multiple transports, doesn't it? 

> Is it true that 192.168.248.40 is an IP address on the machine you are running on?

Yes, it is. My current installation includes a SIP proxy which is bound to 192.168.248.40:5060, while the service based on resiprocate library, namely Registrar, uses 192.168.248.39. That's why I didn't add a transport on 192.168.248.40.
In addition, both addresses are configured to the same network interface, where the second one is an alias. 
```
eth1: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.168.248.39  netmask 255.255.255.0  broadcast 192.168.248.255
        ether 08:00:27:c9:91:80  txqueuelen 1000  (Ethernet)
        ...

eth1:extra: flags=4163<UP,BROADCAST,RUNNING,MULTICAST>  mtu 1500
        inet 192.168.248.40  netmask 255.255.255.255  broadcast 192.168.248.255
        ether 08:00:27:c9:91:80  txqueuelen 1000  (Ethernet)
```
Nevertheless, I believe the configuration without the network alias will have the same issue.

How do you think, which way is better to resolve my problem with your look at the design?

Thanks in advance.

Best regards,
Vyacheslav

vak-oskh:
Hello @sgodin,

I've checked your suggestion. It works for the previously discussed case. 
However, I found another transmitting issue due to my specific network configuration.

The patch becomes more complex and complex to resolve issues for my particular network settings.
Therefore, I've decided to decline this pull request to avoid unclear implementation of `TransportSelector::transmit()` in the master branch. 
I believe this patch could be restored and reviewed if the issue is revealed with the same network settings pattern for somebody else but not today.

By the way, I moved the tests for TransportSelector to a separate pull request. I think they are useful. See details here - https://github.com/resiprocate/resiprocate/pull/364

Thank you for your thoughts and review once more. It is very valuable to me.

Best regards,
Viacheslav

resip/stack/test/CMakeLists.txt
vak-oskh: See the restored tests here - #364

PR #366, opened 2024-08-01
resip/stack: improve p_received handling  without value

vak-oskh:
The Via parameter "received" is very similar to "rport". Neither RFC3581 nor RFC3261 exactly describes the "received" parameter with no value, while does not reject this possibility. Some User Agents treat both parameters similarly and send an empty "received" parameter in a request.

### DIFF
--- a/resip/stack/ReceivedParameter.cxx
+++ b/resip/stack/ReceivedParameter.cxx
@@ -1,0 +1,75 @@
+#include "resip/stack/ReceivedParameter.hxx"
+#include "resip/stack/Symbols.hxx"
+#include "rutil/ParseBuffer.hxx"
+#include "rutil/Logger.hxx"
+
+using namespace resip;
+
+#define RESIPROCATE_SUBSYSTEM resip::Subsystem::SIP
+
+ReceivedParameter::ReceivedParameter(ParameterTypes::Type type,
+                                     ParseBuffer& pb,
+                                     const std::bitset<256>& terminators) :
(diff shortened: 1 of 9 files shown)


resip/stack/ReceivedParameter.cxx
```diff
@@ -0,0 +1,77 @@
...
+      try
+      {
+         pb.skipChar();
+         pb.skipWhitespace();
+
+         // Handle cases such as ";received="
+         if(terminators[(unsigned char)(*pb.position())])
+         {
+            ErrLog(<< "Empty value in string-type parameter.");
+            DebugLog(<< "Define p_received with no value");
```
sgodin, line 29: What is this DebugLog?  There is already and ErrLog, what purpose does the DebugLog serve?  What is the text of the debug log intending to communicate?  In particular what did you mean by "Define"?

sgodin:
Just the one logging comment - otherwise looks sane to me.  Note:  I'm on vacation starting now and all next week.  So I likely won't be responsive during this.

resip/stack/ReceivedParameter.cxx
vak-oskh: Sorry, it was my investigation/testing log. I have removed it.

vak-oskh:
> Just the one logging comment - otherwise looks sane to me. Note: I'm on vacation starting now and all next week. So I likely won't be responsive during this.

Thank you for the notice. Have a great vacation!

PR #367, opened 2024-08-01
resip/stack: fix and improve TLS transports selection

vak-oskh:
The current pull request includes two changes:
- fix TLS transports lookup if only IP addresses are used. All TLS related transports independently of domain name availability are added to `TransportSelector::mTlsTransports`. In accordance, any TLS transport lookup must be done there as well.
- improve TLS transports searching if a domain name is used for TLS transport whether the same IP address with different ports, such as service listen to different ports, or different IP addresses in case the wildcard domain is used. It takes effect for TLS transports with IP addresses only as well.

For example,
- `TLS:192.168.1.1:5060:sip.example.com, TLS:192.168.1.1:5080:sip.example.com`
- `TLS:192.168.1.1:5060:*.example.com, TLS:192.168.1.2:5060:*.example.com`
- `TLS:192.168.1.1:5060, TLS:192.168.1.1:5080, TLS:192.168.1.2:5060`

### DIFF
--- a/resip/stack/SipStack.cxx
+++ b/resip/stack/SipStack.cxx
@@ -548,8 +548,7 @@ SipStack::addTransport(std::unique_ptr<Transport> transport)
    }
    else
    {
-      tuple.setTargetDomain(transport->tlsDomain());
-      TransportSelector::TlsTransportKey tlsKey(tuple);
+      TransportSelector::TlsTransportKey tlsKey(transport->tlsDomain(), tuple);
       if(mSecureTransports.count(tlsKey) == 0)
       {
          // All is good - assign key to transport then add to mNonSecureTransports list
@@ -560,7 +559,8 @@ SipStack::addTransport(std::unique_ptr<Transport> transport)
       else
       {
          // Secure transport collision with existing transport
-         ErrLog(<< "Failed to add secure transport, transport with similar properties already exists: " << tuple);
+         ErrLog(<< "Failed to add secure transport, transport with similar properties already exists: "
+                << tuple << " (domain=" << transport->tlsDomain() << ")");
          throw Transport::Exception("Failed to add secure transport, transport with similar properties already exists.", __FILE__,__LINE__);
          return;
       }
--- a/resip/stack/TransportSelector.cxx
+++ b/resip/stack/TransportSelector.cxx
@@ -1589,9 +1589,7 @@ TransportSelector::findTransportBySource(Tuple& search, const SipMessage* msg) c
 {
    DebugLog(<< "findTransportBySource(" << search << ")");
 
-   if(msg && 
-      !msg->getTlsDomain().empty() && 
-      isSecure(search.getType()))
+   if(msg && isSecure(search.getType()))
    {
       // We should not be willing to attempt sending on a TLS/DTLS/WSS transport 
       // that does not have the cert we're attempting to use, even if the 
@@ -1683,34 +1681,41 @@ Transport*
 TransportSelector::findTlsTransport(const Data& domainname, TransportType type, IpVersion version) const
 {
    resip_assert(isSecure(type));
-   DebugLog(<< "Searching for " << toData(type) << " transport for domain='"
-                  << domainname << "'" << " have " << mTlsTransports.size());
 
    if (domainname == Data::Empty)
    {
-      for(TlsTransportMap::const_iterator i=mTlsTransports.begin(); i != mTlsTransports.end();++i)
+      DebugLog(<< "Searching for " << toData(type) << " transport without domain."
+               << " Secure transports list size = " << mTlsTransports.size());
+
+      for(const auto& tlsTransport : mTlsTransports)
       {
-         if(i->first.mTuple.getType() == type && i->first.mTuple.ipVersion() == version)
+         const TlsTransportKey &key = tlsTransport.first;
+
+         if (key.mTuple.getType() == type &&
+             key.mTuple.ipVersion() == version)
          {
-            DebugLog(<<"Found a default transport.");
-            return i->second;
+            DebugLog(<< "findTlsTransport (exact match) => " << *(tlsTransport.second));
+            return tlsTransport.second;
          }
       }
    }
    else
    {
-      TlsTransportKey key(domainname, type, version);
-      TlsTransportMap::const_iterator i=mTlsTransports.find(key);
+      DebugLog(<< "Searching for " << toData(type) << " transport for domain='"
+               << domainname << "'. Secure transports list size = " << mTlsTransports.size());
 
-      if(i!=mTlsTransports.end())
+      TlsTransportKey key(domainname, type, version);
+      const auto& i = mTlsTransports.find(key);
+
+      if(i != mTlsTransports.end())
       {
-         DebugLog(<< "Found a transport.");
+         DebugLog(<< "findTlsTransport (domain match) => " << *(i->second));
          return i->second;
       }
    }
 
-   DebugLog(<<"No transport found.");
-   return 0;
+   DebugLog(<< "No TLS transport found");
+   return nullptr;
 }
 
 unsigned int
--- a/resip/stack/TransportSelector.cxx
+++ b/resip/stack/TransportSelector.cxx
@@ -67,6 +67,37 @@ using namespace resip;
 
 #define RESIPROCATE_SUBSYSTEM Subsystem::TRANSPORT
 
+TransportSelector::TlsTransportKey::TlsTransportKey(const resip::Data& domainName, const resip::Tuple& tuple) :
+   mTuple(tuple)
+{
+   mTuple.setTargetDomain(domainName);
+}
+
+TransportSelector::TlsTransportKey::TlsTransportKey(const resip::Data& domainName, resip::TransportType type, resip::IpVersion version) :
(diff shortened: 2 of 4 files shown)


vak-oskh:
Hello @sgodin,

I would like to ask you to take a look at my changes if you have time.
I am going to use them and it will be nice to avoid extra commits in my local resiprocate repository if it is accepted.

Thank you in advance.

Best regards,
Vyacheslav

sgodin:
I'm still away on vacation, and will be playing catch up next week.  I'll try to have a look sometime in the next few weeks.

resip/stack/TransportSelector.hxx
```diff
@@ -181,16 +189,27 @@ class TransportSelector
...
+                     }
+                     else if(mTuple.ipVersion() == rhs.mTuple.ipVersion())
+                     {
+                        if (mTuple.getPort() < rhs.mTuple.getPort())
+                        {
+                           return true;
+                        }
+                        else if (mTuple.getPort() == rhs.mTuple.getPort())
+                        {
+                           return mTuple.presentationFormat() < rhs.mTuple.presentationFormat();
```
sgodin, line 204: What is the purpose of the presentationFormat comparison?  It does an inet_ntop- it could be quite expensive if it get's down to this check.

resip/stack/TransportSelector.cxx
```diff
@@ -1547,15 +1547,14 @@ TransportSelector::findTransportBySource(Tuple& search, const SipMessage* msg) c
 {
    DebugLog(<< "findTransportBySource(" << search << ")");
 
-   if(msg && 
-      !msg->getTlsDomain().empty() && 
-      isSecure(search.getType()))
+   if(msg && isSecure(search.getType()))
    {
       // We should not be willing to attempt sending on a TLS/DTLS/WSS transport
```
sgodin, line 1552: I assume adding a TLS transport without a domain is only valid to do if resip is going to as a TLS client only, and will not accept inbound TLS connections.  Is that correct?

Assuming so, this change makes the comment text invalid, and it should be updated.

resip/stack/TransportSelector.cxx
```diff
@@ -1547,15 +1547,14 @@ TransportSelector::findTransportBySource(Tuple& search, const SipMessage* msg) c
...
-      isSecure(search.getType()))
+   if(msg && isSecure(search.getType()))
    {
       // We should not be willing to attempt sending on a TLS/DTLS/WSS transport 
       // that does not have the cert we're attempting to use, even if the 
       // IP/port/proto match. If we have not specified which identity we want
       // to use, then proceed with the code below.
-      return findTlsTransport(msg->getTlsDomain(),search.getType(),search.ipVersion());
+      return findTlsTransport(msg->getTlsDomain(), search.getType(), search.ipVersion(),
+                              search.presentationFormat(), search.getPort());
```
sgodin, line 1557: I'm debating if using search.getSockaddr vs presentationFormat would be better / more efficient.  There is sample sockaddr compare code in Tuple::AnyPortCompare::operator().  Have you given this any thought?

From what I can gleam - the comparison of the presentation format string may have similar cost to doing a sockaddr compare, however the creation of the presentationFormat string bears some cost (inet_ntop), that may indicate sockaddr would be more efficient here.

resip/stack/TransportSelector.hxx
vak-oskh: It is a good question. I am going to analyze a better approach.
The purpose is to find a proper TLS transport in case the service is configured by more than one TLS transport.
For example,
```
transportSelector.addTransport("192.168.1.1", 5060, V4, TLS);
transportSelector.addTransport("192.168.1.2", 5060, V4, TLS);
```

resip/stack/TransportSelector.cxx
vak-oskh: Actually, we are talking about the transport selection which is used by TransportSelector::transmit(). It means an outgoing SIP message sending but not TLS handshake.
That's why this code can be applied to a service TLS listening transports.

Let's imagine Registrar listens to `TLS:1.2.3.4:5061` where no domain is assigned.
The service has the functionality to send SIP requests, like SIP OPTIONS for NAT pings and UA is registered via TLS. 
Of course, there must be an SSL certificate with the particular IP address to create a TLS connection;
however, transport selection seems used to find which interface the service is going to create/re-use the connection.

To be honest, I do not understand totally the comment "We should not be willing ... then proceed with the code below.". I expect that a service chooses the transport for sending SIP requests and only after that TLS connection is created/re-used. It looks like two separate layers. Correct me if I am wrong.

On the other hand, I removed the condition `!msg->getTlsDomain().empty()` as a TLS transport is added in a separate container and it doesn't work for such selections, see  https://github.com/resiprocate/resiprocate/blob/master/resip/stack/TransportSelector.cxx#L253

resip/stack/TransportSelector.cxx
vak-oskh: I think you are right. I am going to update it today.
To be honest, I missed this point. Thank you for the advice.

resip/stack/TransportSelector.cxx
vak-oskh: In addition, if to talk about "a TLS transport without a domain", I believe it is not so popular to bind TLS transport without a domain on the Internet. However, it is not forbidden.
For example, you have a small office with a SIP server installation and don't want to purchase a domain. It should work well.

resip/stack/TransportSelector.hxx
vak-oskh: The pull request has been updated. The tuple address comparison as `resip::Data` was replaced with the binary one, which is present in `resip::Tuple::operator<()`

vak-oskh:
Hello @sgodin, 
Could you look at my changes when you have some free time?

Thank you in advance.

Best regards,
Vyacheslav

resip/stack/TransportSelector.cxx
```diff
@@ -1638,37 +1633,47 @@ TransportSelector::findTransportBySource(Tuple& search, const SipMessage* msg) c
...
-         if(i->first.mTuple.getType() == type && i->first.mTuple.ipVersion() == version)
+         const TlsTransportKey &key = tlsTransport.first;
+
+         if (key.mTuple == search)  // Tuple does not compare domain names by default.
          {
-            DebugLog(<<"Found a default transport.");
-            return i->second;
+            DebugLog(<< "findTlsTransport (exact match) => " << *(tlsTransport.second));
+            return tlsTransport.second;
          }
```
sgodin, line 1657: If we fail to match with Tuple ==, should we fallback to the old code matching and just try to find match on type and ipversion?  We could keep track of the "best" match as we iterate through the list.  Just worried about making sure we are backwards compatible here.

resip/stack/SipStack.cxx
```diff
@@ -548,8 +548,7 @@ SipStack::addTransport(std::unique_ptr<Transport> transport)
    }
    else
    {
-      tuple.setTargetDomain(transport->tlsDomain());
-      TransportSelector::TlsTransportKey tlsKey(tuple);
+      TransportSelector::TlsTransportKey tlsKey(transport->tlsDomain(), tuple);
       if(mSecureTransports.count(tlsKey) == 0)
       {
          // All is good - assign key to transport then add to mNonSecureTransports list
```
sgodin, line 554: tuple is logged out on line 562 - it will no longer have the targetDomain set on it, we should probably add the domain to that log statement

sgodin:
> Hello @sgodin, Could you look at my changes when you have some free time?
> 
> Thank you in advance.
> 
> Best regards, Vyacheslav

@vak-oskh  Sorry for the delay - this is looking pretty good to me.  Just a few dangling comments.  Thanks for your patience and contribution.

Scott

sgodin:
Hi @vak-oskh - is this ready for another look?

vak-oskh:
> Hi @vak-oskh - is this ready for another look?

Hello @sgodin, yes is it.
Thank you in advance.

resip/stack/TransportSelector.cxx
sgodin: I think the comment is saying that if the sipMessage has a TlsDomain set on it, then we should only choose the TLS transport for that domain to send on (ie: the transport that has that domain cert set on it that matches the domain set in the SipMessage).   

That being said I'm not sure offhand how, when or why the TlsDomain get's set on a SipMessage.

sgodin:
@vak-oskh Thanks for all your patience with this.  Since this brings in reasonably large changes, I wanted to make sure we didn't break anything.  I don't see any other theoretical issues, so I think it's time to push this through and see if the community notices any issues with it.  

Thanks again!
Scott

vak-oskh:
Hello @sgodin,
thank you for the review.

I see that you have closed the PR; however, the changes have not been merged.
It looks like a mistake or I didn't understand your last message.

Thank you in advance!
Vyacheslav

sgodin:
> Hello @sgodin,
> 
> thank you for the review.
> 
> 
> 
> I see that you have closed the PR; however, the changes have not been merged.
> 
> It looks like a mistake or I didn't understand your last message.
> 
> 
> 
> Thank you in advance!
> 
> Vyacheslav

Oops, I'll fix it.  Thanks.

PR #368, opened 2024-08-14
repro: restore the behavior of ResponseContext when 503 is received to follow RFC 3261

vak-oskh:
[Commit 6cf081f](https://github.com/resiprocate/resiprocate/commit/6cf081f9cbcfbfdb2597b18f81c267b79f9859bb) changed the mentioned logic in ResponseContext a long time ago.

The current patch restores the behaviour to respond with a 500 code when a 503 is received, following [RFC 3261](https://datatracker.ietf.org/doc/html/rfc3261#page-111).
In addition, the possibility to implement your custom logic for `repro::ResponseContext::forwardBestResponse()` by the derived class has been added.

The flow to add the custom logic is the following:
* `repro::Proxy::setRequestContextFactory(CustomRequestContextFactory)`
* `CustomRequestContextFactory::createRequestContext() -> CustomRequestContext`
* `CustomRequestContext().mResponseContext.reset(new CustomResponseContext(*this))`
* `CustomResponseContext::forwardBestResponse()` implements your logic

### DIFF
--- a/repro/ResponseContext.cxx
+++ b/repro/ResponseContext.cxx
@@ -1533,8 +1533,8 @@ ResponseContext::forwardBestResponse()
    
    if(mBestResponse.header(h_StatusLine).statusCode() == 503)
    {
-      //See RFC 3261 sec 16.7, page 110, paragraph 2
-      mBestResponse.header(h_StatusLine).statusCode() = 480;
+      // See RFC 3261 sec 16.7, page 111, paragraph 2
+      mBestResponse.header(h_StatusLine).statusCode() = 500;
    }
 
    if(mBestResponse.header(h_StatusLine).statusCode() == 408 &&
--- a/repro/RequestContext.cxx
+++ b/repro/RequestContext.cxx
@@ -46,7 +46,7 @@ RequestContext::RequestContext(Proxy& proxy,
    mTransactionCount(1),
    mProxy(proxy),
    mTopRouteFlowTupleSet(false),
-   mResponseContext(*this),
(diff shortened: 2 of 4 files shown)


repro/ResponseContext.cxx
```diff
@@ -1533,8 +1533,8 @@ ResponseContext::forwardBestResponse()
    
    if(mBestResponse.header(h_StatusLine).statusCode() == 503)
    {
-      //See RFC 3261 sec 16.7, page 110, paragraph 2
-      mBestResponse.header(h_StatusLine).statusCode() = 480;
+      // See RFC 3261 sec 16.7, page 111, paragraph 2
+      mBestResponse.header(h_StatusLine).statusCode() = 500;
```
sgodin, line 1537: FYI - I have reached out to @docfaraday to see if he can remember back 18 years ago, as to why this change was made in the first place. :)

vak-oskh: I believe my solution should work without changes to follow RFC and with some extra changes with a derived class if you want some custom SIP code here.

sgodin: This was docfaraday's response:
     I'm not really sure. However, 500 makes almost no sense to me. It
gives no hint at all what went wrong, or where (did it fail at the
proxy? downstream?). That might be why I changed it? 480 at least would
suggest an appropriate course of action for the human caller (ie; try
again later), but it has its own problems too though. I don't know that
there actually is a good option.

I think we are good to proceed with the change to follow the RFC, and if someone wants to override then they can.

PR #369, opened 2024-08-22
build: add CMake flag to exclude tests

vak-oskh:
This option might be useful for release builds where tests are not required. 
It reduces building time and makes the installation directory clean.

Just use `-DBUILD_TESTING=OFF` to build the resiprocate library without unit and manual tests.


sgodin:
Looks good - thanks!

PR #375, opened 2024-10-08
resip/stack: improve duplicate parameters handling in ParserCategory

mykhailopopivniak:
RFC 3261 prohibits any given parameter-name to appear more than once in URIs and header field values. Skip duplicate parameter-names during the parsing procedure.


sgodin:
This change looks good to me!  Thank you for the contribution and including the additional unit tests.

PR #378, opened 2024-10-18
Allow ACK to be handled in the SentReInvite state

Darren-Wade:
We found a problem interworking with an Oracle SBC.

We had received an INVITE and responded with the 200OK (a=inactive) indicating we are not ready for audio. 
But before receiving the the ACK, we had sent a Re-INVITE (a=recvonly) to request the audio.

This Re-INVITE moved the session into the 'SentReInvite' state.
So the ACK arrives into InviteSession::dispatchSentReinvite, where there was no case to handle it.
This resulted in the 200OK re-transmit timer not being cancelled.

This change adds a case to handle the ACK and cancel the 200OK re-transmit timer, when in the 'SentReInvite' state

### DIFF
--- a/resip/dum/InviteSession.cxx
+++ b/resip/dum/InviteSession.cxx
@@ -1729,6 +1729,11 @@ InviteSession::dispatchSentReinvite(const SipMessage& msg)
          handler->onOfferRejected(getSessionHandle(), &msg);
          break;
 
+      case OnAck:
+      case OnAckAnswer:
+         mCurrentRetransmit200 = 0; //stop the 200 retransmit timer
+      break;
+
       default:
          dispatchOthers(msg);
          break;
--- a/resip/dum/InviteSession.cxx
+++ b/resip/dum/InviteSession.cxx
@@ -2148,6 +2148,7 @@ InviteSession::dispatchOthers(const SipMessage& msg)
 {
    // handle OnGeneralFailure
    // handle OnRedirect
+   InviteSessionHandler* handler = mDum.mInviteSessionHandler;
 
    switch (msg.header(h_CSeq).method())
    {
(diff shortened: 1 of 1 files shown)


sgodin:
@Darren-Wade thanks for this contribution!

If we look at how an ACK is handled when normally in the connected state (ie: InviteSession::dispatchConnected):

      case OnAck:
      case OnAckAnswer:
         // Not checking for strict equality, since request may have been digest challenged
         if (mLastRemoteSessionModification->header(h_CSeq).sequence() > msg.header(h_CSeq).sequence())
         {
            InfoLog(<< "dropped stale ACK");
         }
         else
         {
            mCurrentRetransmit200 = 0; // stop the 200 retransmit timer
            handler->onAckReceived(getSessionHandle(), msg);
         }
         break;

I'm thinking we should use the same handling in the changes you made to dispatchSentReInvite.  Can you please change your code to do this an confirm the fix still works for you?  This way the onAckReceived callback will still get called and we get verification of the CSeq.

resip/dum/InviteSession.cxx
```diff
@@ -2161,9 +2175,17 @@ InviteSession::dispatchOthers(const SipMessage& msg)
...
+      case ACK:
+         // Not checking for strict equality, since request may have been digest challenged
+         if (mLastRemoteSessionModification->header(h_CSeq).sequence() > msg.header(h_CSeq).sequence())
+         {
+            InfoLog(<< "dropped stale ACK");
+         }
+         else
+         {
+            mCurrentRetransmit200 = 0; // stop the 200 retransmit timer
+         }
```
sgodin, line 2187: Is there a reason you are not calling:
handler->onAckReceived(getSessionHandle(), msg);
here?

Darren-Wade: I was trying to do the minimum in this default method, but I will revert the change from the last commit to make it in line with the code in dispatchSentReinvite.

sgodin: I think it's good to leave the code in dispatchOthers - but I think it also makes sense to call onAckReceived  :)

Darren-Wade: Now calling onAckRecieived from dispatchOthers.

sgodin:
Thanks @Darren-Wade !

PR #385, opened 2024-10-30
Improve URI comparison logic

mykhailopopivniak:
- Extract repetetive code into separate functions to improve maintainability and extensibility.
- According to RFC 3261 and RFC5627, `gr` parameter should be considered in URI comparison if it appears in both compared URIs.
-  Consider URI parameters in Uri's `operator<` to ensure correct logic, consistent with Uri's `operator==`.

### DIFF
--- a/resip/stack/Uri.cxx
+++ b/resip/stack/Uri.cxx
@@ -450,6 +450,40 @@ bool Uri::compareUriParametersEqual(Parameter* param1, Parameter* param2)
    }
 }
 
+bool Uri::compareUriParametersLessThan(Parameter* param1, Parameter* param2)
+{
+   if (!param1 || !param2)
+   {
+      return false;
+   }
+
+   switch (param1->getType()) {
+      case ParameterTypes::user:
+      case ParameterTypes::method:
+      case ParameterTypes::maddr:
+      case ParameterTypes::transport:
+         return isLessThanNoCase(dynamic_cast<DataParameter*>(param1)->value(),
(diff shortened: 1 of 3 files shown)


mykhailopopivniak:
Hi, @sgodin. Could you review this PR when you have time? It would be great to get this included in the next release. Let me know if there's anything that needs adjustment. Thanks!

sgodin:
@mykhailopopivniak Will do!  I want to make sure all TFM tests pass first, then I'll merge in and make sure no tests are broken.

resip/stack/Uri.cxx
```diff
@@ -748,7 +703,95 @@ Uri::operator<(const Uri& other) const
...
+      unB.insert(*i);
+   }
+#else
+   ParameterList unA = mUnknownParameters;
+   ParameterList unB = other.mUnknownParameters;
+
+   sort(unA.begin(), unA.end(), orderUnknown);
+   sort(unB.begin(), unB.end(), orderUnknown);
+#endif
+
```
sgodin, line 759: Can you please add some comments in the code explaining what this section of code is doing?  ie:  Why build up 2 new sets?  Why is code different for different platforms?  And a general explanation of the logic below.  :)  Thank you in advance.

mykhailopopivniak: Added comments and renamed some variables to make it easier to understand.
This part of the code mirrors the one from `Uri::operator==`.

PR #386, opened 2024-11-06
Improve the implementation of resip::Data interoperability with std::string and std::string_view

mykhailopopivniak:
Add a complete set of comparison operators (<, >, <=, >=) to resip::Data for interoperability with std::string and std::string_view.
Other small improvements.


sgodin:
@mykhailopopivniak Thanks for this, looks great!

PR #389, opened 2024-12-20
fix: detect std::string_view availability at configure time

gjasny:
Allowing the binary layout of a class to differ between library compilation and usage is a pretty bad idea and leads to hard-to-find memory corruption bugs.

Instead of detecting support for std::string_view at compile time we now check for it at configure time and record the decision into the config.h file. That way both: library and users use the same class layout.

The downside of this approach is that the checked-in Visual Studio project files don't support any such detection and thus cannot use the feature. But in the long-term those files should vanish and CMake be used exclusively.


sgodin:
Makes sense to me.  Is this ready to merge?  I'm not worried about the checked in Visual Studio Project files - since those are deprecated, and will eventually go away (as you rightfully pointed out).

PR #393, opened 2025-01-05
Fix windows testStack test

gjasny:
On Systems where `getLocalHostName` only returns a hostname without domain, the `testStack` test asserts and hangs.

### DIFF
--- a/resip/stack/test/testStack.cxx
+++ b/resip/stack/test/testStack.cxx
@@ -603,7 +603,7 @@ main(int argc, char* argv[])
    const char* logLevel = "WARNING";
    const char* proto = "tcp";
    const char* bindAddr = "127.0.0.1";
-   const char* domain = "";
+   const char* domain = "resiprocate.test";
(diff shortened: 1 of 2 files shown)


resip/stack/test/testStack.cxx
```diff
@@ -603,7 +603,7 @@ main(int argc, char* argv[])
    const char* logLevel = "WARNING";
    const char* proto = "tcp";
    const char* bindAddr = "127.0.0.1";
-   const char* domain = "";
+   const char* domain = "resiprocate.test";
```
gjasny, line 606: According to [RFC2602](https://www.rfc-editor.org/rfc/rfc2606.html#page-2) the `test` TLD is reserved.

sgodin:
Hi @gjasny I've incorporated these fixes into my latest windows-ci branch.  I will close this PR as the fixes are included when I merged that branch.  Thank you for investigating this!

PR #400, opened 2025-03-02
Some small optimizations 

portsip:
- Replace set with unordered_set in BasicDomainMatcher and ExtendedDomainMatcher.
- Use find() instead of count() for element existence checks in STL containers.
- Optimize locking in DnsInterface::isSupported().
- Early return in ExtendedDomainMatcher::isMyDomain() when mDomainSuffixList is empty, preventing unnecessary domain case conversions.


sgodin:
Looks good to me - thanks for the contribution!

PR #405, opened 2025-04-20
Optimize lock scopes for improved efficiency

portsip:


resip/dum/InMemorySyncRegDb.cxx
```diff
@@ -253,10 +254,10 @@ InMemorySyncRegDb::aorIsRegistered(const Uri& aor, uint64_t* maxExpires)
 void
 InMemorySyncRegDb::lockRecord(const Uri& aor)
 {
-   Lock g2(mLockedRecordsMutex);
-
    DebugLog(<< "InMemorySyncRegDb::lockRecord:  aor=" << aor << " threadid=" << ThreadIf::selfId());
```
sgodin, line 258: I think you should reverse this logging change, it could be confusing.  If two threads call lockRecord, both will log "lockRecord" but only one will get in first.  It's better to log after the lock is grabbed, so it only logs, when we know the record actually get's locked.  This comment applies to unlockRecord as well, although there should never be two threads calling this method at the same time.

portsip: Thanks for pointing that out. 
It's fixed now.

PR #412, opened 2025-07-01
Add ServerRegistration::reject overload that takes a specific SipMessage

bwalden:


resip/dum/ServerRegistration.cxx
```diff
@@ -143,6 +143,33 @@ ServerRegistration::accept(int statusCode)
...
+ServerRegistration::reject(SipMessage& failure)
+{
+   failure.remove(h_Contacts);
+
+   InfoLog( << "rejected a registration " << mAor << " with statusCode=" << failure.header(h_StatusLine).statusCode() );
+
+   // First, we roll back the contact database to
+   // the state it was before the registration request.
+
+   // Async processing hasn't actually updated the database yet, so no need to roll back.
```
sgodin, line 156: There is bunch of code duplication here with the existing reject method.  Would be nice to consolidate these two methods more.  For example the existing Reject(int statusCode) method could call this new one after it has created the SipMessage to send.  Similar to the way the 2 accept methods work, something like:

void
ServerRegistration::reject(int statusCode)
{
   SipMessage failure;
   mDum.makeResponse(failure, mRequest, statusCode);
   reject(failure);
}

bwalden: Agreed. I've updated reject(int) to call reject(SipMessage&).

resip/dum/ServerRegistration.cxx
```diff
@@ -143,6 +143,33 @@ ServerRegistration::accept(int statusCode)
    accept(success);
 }
 
+void
+ServerRegistration::reject(SipMessage& failure)
+{
+   failure.remove(h_Contacts);
```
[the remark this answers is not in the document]
bwalden: These validations were also missing from the existing accept(SipMessage&). I have added them there too.

PR #418, opened 2025-08-05
Allow to reject calls with a body content.

jmariasc:
3GPP TS 24 229 describes the possibility of including a multipart/mixed body in 486 (Busy Here), 600 (Busy Everywhere) or 603 (Decline) response to the INVITE request.

### DIFF
--- a/resip/dum/ServerInviteSession.cxx
+++ b/resip/dum/ServerInviteSession.cxx
@@ -635,6 +635,14 @@ ServerInviteSession::reject(int code, WarningCategory *warning)
          // we should send 200PRACK
          auto response = std::make_shared<SipMessage>();
          mDialog.makeResponse(*response, mFirstRequest, code);
+         if (mProposedLocalOfferAnswer)
+         {
+            setOfferAnswer(*response, mProposedLocalOfferAnswer.get());
+         }
+         else if (mCurrentLocalOfferAnswer)
+         {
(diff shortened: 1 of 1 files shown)


sgodin:
This change interacts with the offer/answer state machine which is not really appropriate for adding a body to the BYE message.  You could use some of the message decoration mechanisms DUM offers instead:  https://github.com/resiprocate/resiprocate/wiki/Modifying-Decorating-messages-sent

The onReadyToSend override is probably the easiest for what you want to do.

jmariasc:
Thanks for checking this out. I'm familiar with the onReadyToSend override, and I'm already using it to add headers to methods and responses in an IMS environment, but I thought that to add a complete body part without having called accept(), it would be more appropriate to use the provideOffer/provideAnswer mechanism just before calling reject(), considering that the offer/response state machine ends right there, entering the Terminated state. Maybe narrowing the call to setOfferAnswer() into the UAS_Offer and UAS_NoOffer states would be more appropriate.

jmariasc:
I also think a good solution would be to add a reject method with a third parameter. This method would only affect the "ServerInviteSession" class. If you agreed I could submit a pull request.

sgodin:
I agree.  You will also need to modify InviteSession as well, since the ServerInviteSession method calls into the base class InviteSession for handling the reject call for states after the initial server side call setup.

PR #420, opened 2025-08-11
Added a reject method with a Contents object.

jmariasc:
This method allows you to include a body in a reject response without affecting the DUM offer/answer state machine.

### DIFF
--- a/resip/dum/ClientInviteSession.cxx
+++ b/resip/dum/ClientInviteSession.cxx
@@ -266,6 +266,11 @@ ClientInviteSession::reject (int statusCode, WarningCategory *warning)
       {
          auto response = std::make_shared<SipMessage>();
          mDialog.makeResponse(*response, *mLastRemoteSessionModification, statusCode);
+         // In case we wanted to reject the request by including content
+         if (mProposedLocalReject)
+         {
+            response->setContents(mProposedLocalReject.get());
+         }
          if(warning)
          {
             response->header(h_Warnings).push_back(*warning);
@@ -303,6 +308,13 @@ ClientInviteSession::reject (int statusCode, WarningCategory *warning)
          InviteSession::reject(statusCode, warning);
          break;
    }
+}
+
+void
+ClientInviteSession::reject (int code, const Contents& rejectBody, WarningCategory *warning)
+{
+   mProposedLocalReject = std::unique_ptr<Contents>(rejectBody.clone());
+   reject(code, warning);
 }
 
 void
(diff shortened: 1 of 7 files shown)


sgodin:
I don't see a reason to add the additional state storage: 
std::unique_ptr<Contents> mProposedLocalReject;

I think you should instead do the following:
virtual void reject(int statusCode, const Contents* contents, WarningCategory* warning = 0);
^^^ Modify the existing reject code to do what you want, ie: if contents != 0 add to body

virtual void reject(int statusCode, WarningCategory *warning = 0);
^^^ Add back old method signature for ABI compatibility that calls: reject(statusCode, 0, warning);

Note:  There is also a rejectCommand method that should/could be extended to offer the same functionality in a thread safe API.  A similar approach can be used.

sgodin:
Those changes look great to me - thank you for the contribution.  Note:  Even though InviteSession::rejectCommand is not used in the resip code base, it is used in some resiprocate users applications, so it's a good idea to extend it as well for consistency.  Thank you for doing that as well.  I will start the CI build now on this branch to ensure everything builds and unit tests still pass, then I can approve this PR.

jmariasc:
This was due to the reject method overload. The overload requires an explicit declaration of the method to be bound in boost::bind. Fixed in TestInviteSession.cxx and TFM built successfully. I also fixed a little typo in InviteSession.cxx

jmariasc:
dumTests failed due a weird loop:

```
(gdb) bt
...
#6  0x00007ffff6c50217 in resip::ServerInviteSession::reject (this=0x7fffdc4a4640, code=488, contents=0x0, warning=0x0) at /opt/resiprocate/resip/dum/ServerInviteSession.cxx:609
#7  0x00007ffff6c1c7cc in resip::InviteSession::reject (this=0x7fffdc4a4640, statusCode=488, warning=0x0) at /opt/resiprocate/resip/dum/InviteSession.cxx:814
#8  0x00007ffff6c5055d in resip::ServerInviteSession::reject (this=0x7fffdc4a4640, code=488, contents=0x0, warning=0x0) at /opt/resiprocate/resip/dum/ServerInviteSession.cxx:675
#9  0x00007ffff6c1c7cc in resip::InviteSession::reject (this=0x7fffdc4a4640, statusCode=488, warning=0x0) at /opt/resiprocate/resip/dum/InviteSession.cxx:814
...

```
It seems that the mixing of overriding and overloading of the ServerInviteSession::reject method is not working as expected. I'll check internally and come back to you when it be fixed. Thank you.

jmariasc:
In the end it ended up being a silly error by invoking the wrong overloaded method. It builds and passes the tests successfully now: https://github.com/jmariasc/resiprocate/actions/runs/16950346035

PR #426, opened 2025-09-05
Update for Asio 1.34 and later compatibility

Lastique:
This updates usage of Asio for compatibility with version 1.34 and newer, where a number of previously deprecated APIs were removed.

### DIFF
--- a/reTurn/AsyncTcpSocketBase.cxx
+++ b/reTurn/AsyncTcpSocketBase.cxx
@@ -15,7 +15,7 @@ using namespace std;
 
 namespace reTurn {
 
-AsyncTcpSocketBase::AsyncTcpSocketBase(asio::io_service& ioService) 
+AsyncTcpSocketBase::AsyncTcpSocketBase(asio::io_context& ioService) 
  : AsyncSocketBase(ioService),
    mSocket(ioService), 
    mResolver(ioService)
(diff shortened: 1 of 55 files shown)


reTurn/AsyncTcpSocketBase.cxx
```diff
@@ -76,26 +75,17 @@ AsyncTcpSocketBase::handleTcpResolve(const asio::error_code& ec,
...
       // The connection was successful.
       mConnected = true;
-      mConnectedAddress = endpoint_iterator->endpoint().address();
-      mConnectedPort = endpoint_iterator->endpoint().port();
+      mConnectedAddress = endpoint.address();
+      mConnectedPort = endpoint.port();
 
       onConnectSuccess();
    }
-   else if (++endpoint_iterator != asio::ip::tcp::resolver::iterator())
```
sgodin, line 90: Was this else block intentionally removed?

reTurn/AsyncTlsSocketBase.cxx
```diff
@@ -85,23 +84,14 @@ AsyncTlsSocketBase::handleTcpResolve(const asio::error_code& ec,
...
+                                  const asio::ip::tcp::endpoint& endpoint)
 {
    if (!ec)
    {
       // The connection was successful - now do handshake.
       mSocket.async_handshake(asio::ssl::stream_base::client, 
                               std::bind(&AsyncSocketBase::handleClientHandshake, shared_from_this(), 
-                                          std::placeholders::_1, endpoint_iterator));
-   }
-   else if (++endpoint_iterator != asio::ip::tcp::resolver::iterator())
```
sgodin, line 97: Same here?

sgodin:
Thank you for this!  FYI: We have a drop of asio in the contrib folder, it is used for Windows builds, and is version 1.18.1.  It seems the newer API's you are using in this PR must be supported in 1.18.1 as well or the Windows CI build would have failed.  Eitherway, it probably makes sense to update that as well.  I will look into that.

reTurn/AsyncTcpSocketBase.cxx
Lastique: Yes. It restarted the `async_connect` to another remote address after incrementing the iterator. This is handled by the standalone `asio::async_connect` algorithm, which takes a list of endpoints and passes the endpoint that got connected to the callback.

PR #427, opened 2025-09-07
Reduce log level of stats messages

Lastique:
Stats messages do not indicate a problem and therefore should not be logged at warning level.

### DIFF
--- a/resip/stack/StatisticsMessage.cxx
+++ b/resip/stack/StatisticsMessage.cxx
@@ -93,7 +93,7 @@ void
 StatisticsMessage::logStats(const resip::Subsystem& subsystem, 
                             const StatisticsMessage::Payload& stats)
 {
-   WarningLog(<< subsystem
+   InfoLog(<< subsystem
               << std::endl
               << stats);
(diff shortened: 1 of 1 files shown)


sgodin:
Logging at warning was intentional.  Assuming that most production systems might want to run logging at Warning level to filter out all the INFO level logs.  Statistics logs can be disabled all together with SipStack::statisticsManagerEnabled().  If you require logging at INFO instead, perhaps we should make a new statistics setting that specifies the log level it will log at.

Lastique:
We don't necessarily want to disable statistics entirely, but we do want to avoid false warnings and errors in logs since they complicate real error analysis.

My understanding is that if one wants to limit logs to warnings and up, plus statistics, he should be able to enable stats at all levels and the rest at warning level and above. You would normally use channels to implement this separation, and libresiprocate has `Subsystem`s as the closest equivalent. You even have the `STATS` subsystem.

So, is it not possible to configure different log level thresholds for different subsystems? Would it not be enough to support the use case you describe?

PR #429, opened 2025-09-09
Add support for session refresh using OPTIONS

Lastique:
When a SIP endpoint does not support timers extension (RFC4028) but supports OPTIONS request (which it should, as this is one of the basic requests defined in RFC3261), using OPTIONS instead of re-INVITE to perform session refresh can be beneficial. OPTIONS is much more lightweight as it doesn't update SDP, which in some cases may disrupt media exchange.

Note that OPTIONS can be sent in- or out-of-dialog. In this implementation we are sending in-dialog OPTIONS to verify that this dialog is still active.

### DIFF
--- a/resip/dum/DialogSet.cxx
+++ b/resip/dum/DialogSet.cxx
@@ -635,6 +635,14 @@ DialogSet::dispatch(const SipMessage& msg)
             }
             break;
 
+         case OPTIONS:
+            if (dialog)
+            {
+               DebugLog(<< "in dialog options request");
+               break;
+            }
+            // fallthrough
+
          default:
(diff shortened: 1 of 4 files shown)


resip/dum/DialogSet.cxx
```diff
@@ -635,6 +635,19 @@ DialogSet::dispatch(const SipMessage& msg)
             }
             break;
 
+         case OPTIONS:
```
sgodin, line 638: You could reduce code duplication here, by only breaking if dialog is not null, thus falling through to the default case if dialog is null.

sgodin:
This looks great - thank you for the submission!

PR #431, opened 2025-09-12
Improve transaction ID uniqueness

Lastique:
Increase the length of the generated transaction IDs by adding a timestamp to reduce the possibility of collisions. Use crypro RNG explicitly so that the generated IDs are less likely to repeat. This reduces the likelihood of duplicate transaction IDs in long-running calls.


sgodin:
From my research RAND_bytes (OpenSSL) is 50-100x slower than rand() pseudo rand generator.  I'm debating if this change should be behind a preprocessor flag.  What are your thoughts?  Were you seeing problems in production with the existing implementation?  If so, can explain what scenario you were seeing issues?  

Thanks!
Scott

PR #436, opened 2025-09-13
Fix usage of `<cctype>` functions

Lastique:
Functions defined in `<cctype>` have undefined behavior if a negative value is passed on input. Input values must always be converted to unsigned integers.

Also, added missing includes.


sgodin:
Thanks @Lastique - I've been a bit busy lately, so I'm slowly chipping away at these PR's.

PR #438, opened 2025-09-17
Add support for iteration over SIP headers

Lastique:
This refactors the `mHeaders` and `mHeaderIndices` containers in `SipMessage` into a dedicated `KnownHeaders` container and exposes this container to users via the new `getRawHeaders()` accessor. This allows users to iterate over the known headers of a SIP message to implement their own logic dependent on header values.

The fundamental design of the container remains the same, although there are some changes. In order to implement iterators, the container elements need to store header type IDs (since, until C++20, iterators' reference type must be an actual reference, and we want the header type to be accessible during iteration). We also reuse this field internally by the container to mark entries as "unused" when a header is removed from the container. "Unused" elements are not exposed through iterators or other public accessors and are reused by the container if the same header is later inserted. Moving this "unused" property into the container elements allowed to reduce the size of the lookup index elements to just one byte per header type. And since index elements are now zero-based, we no longer have to allocate a dummy unused element at index 0.

Since the container now supports proper iteration over headers, replaced instances where it was previously emulated via a loop over all known header types with proper iteration. This should arguably improve performance on average, since it is unlikely that a SIP message contains all or even most of header types supported by libresiprocate.

One side effect of this is that the order of headers observed though iteration (including when `SipMessage` is serialized) may change. The tests where this was the case were updated by reordering the headers.

### DIFF
--- a/resip/stack/SipMessage.cxx
+++ b/resip/stack/SipMessage.cxx
@@ -26,15 +26,110 @@ using namespace std;
 
 #define RESIPROCATE_SUBSYSTEM Subsystem::SIP
 
+/// Clears the container. Does not free `HeaderFieldValueList` objects but clears them and marks as "unused".
+void
+SipMessage::KnownHeaders::clear() noexcept
+{
+   for (iterator it = begin(), e = end(); it != e;)
+   {
+      erase(it++);
+   }
+}
+
+/// Clears the container and invokes the dispose function on every element.
+template<typename Disposer>
+inline void
+SipMessage::KnownHeaders::clearAndDispose(Disposer&& disposer) noexcept
+{
+   for (reference elem : mHeaders)
+   {
+      disposer(elem);
+   }
+
+   mHeaders.clear();
+   mSize = 0;
+   resetIndices();
+}
+
+/// Erases an element. Does not free the `HeaderFieldValueList` object but clears it and marks as "unused".
+void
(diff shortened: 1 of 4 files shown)


Lastique:
> Does this change have any significant impact on the size of a SipMessage in memory?

The lookup index (`mHeaderIndices`) is half the size in the new code but a single-byte `mSize` is added. So, `sizeof(mKnownHeaders) == 136` on my machine (x86-64, gcc 13), while previously it was `sizeof(mHeaders) + sizeof(mHeaderIndices) == 224)`.

The downside is that the new `KnownHeaders::mHeaders` vector contains `HeaderInfo` elements instead of just `HeaderFieldValueList*`, which is twice the size per element (16 bytes vs 8 bytes on 64-bit targets). It no longer has the dummy unused element at offset 0 though. The number of elements in `mHeaders` corresponds to the number of headers in the SIP message, so whether this is a win depends on the number of headers. If I'm counting right, it should be a win up to 6 headers.

> or does it worsen the performance of header parsing and processing?

I did not benchmark, but given that iteration now only processes headers that are actually in the message (plus the "unused" entries, which should typically not exist) rather than just from 0 to `MAX_HEADERS` (96) and that the processing happens in the linear order over `mHeaders` without involving `mHeaderIndices`, I think, the performance should be improved.

PR #440, opened 2025-09-19
fix MSVC compile warning C4819

fnMrRice:
replaced non-ascii quote with ascii quote


sgodin:
@fnMrRice Thanks you for this!

PR #453, opened 2026-01-07
Expose ARES_FLAG_NOCHECKRESP ares flag in resip

szokovacs:
It can be used to force ares to use the hosts file as a fallback when the DNS lookup fails


sgodin:
Hi @szokovacs, thanks for this contribution.  I'd like to understand more how this facilitates fallback to hosts file lookup, can you explain in more detail?

sgodin:
That is great!  I think it would be good to document all that in rutil/dns/ExternalDns.hxx - next to the ReturnAllResults setting

PR #456, opened 2026-02-03
Fix infinite logging loop on socket error ENOTCONN/WSAENOTCONN in Transportfix(transport): handle cleanup for ENOTCONN errors in TcpBaseTransport

MatteoAns:
**Description**: This PR addresses an issue where the stack enters a "tight loop" of infinite logging when a TCP/TLS connection is abruptly disconnected or fails during handshake, specifically returning error 57 (ENOTCONN) on macOS/POSIX or 10057 (WSAENOTCONN) on Windows.

**Root Cause**: The issue stems from the interaction between FdPollImplFdSet::processFdSet and the Transport error handling logic.

When a socket enters a disconnected state, the system selector (select/poll) continuously marks the file descriptor as "ready" to signal an exception or a hung-up state.

FdPollImplFdSet detects this and calls processItem, which invokes the transport's read/write operations.

In Transport.cxx, the error is caught and logged, but the connection is not explicitly terminated or removed from the selector's watch set, since the socket remains "valid" but "not connected" in the selector, the loop repeats immediately, flooding the logs and causing high CPU usage.

### DIFF
--- a/resip/stack/TcpBaseTransport.cxx
+++ b/resip/stack/TcpBaseTransport.cxx
@@ -145,6 +145,7 @@ TcpBaseTransport::buildFdSet( FdSet& fdset)
 
 /**
     Returns 1 if created new connection, -1 if "bad" error,
+    -2 if error which requires cleanup
     and 0 if nothing to do (EWOULDBLOCK)
 **/
 int
@@ -168,6 +169,9 @@ TcpBaseTransport::processListen()
                // !jf! this can not be ready in some cases
                // !kw! this will happen every epoll cycle
                return 0;
+            case ENOTCONN:
+               Transport::error(e);
+               return -2; //require cleanup
             default:
                Transport::error(e);
          }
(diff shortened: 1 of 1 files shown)


sgodin:
Here's my main concern.  You are listening on a TCP/TLS socket, and some disconnection happens during a TCP/TLS handshake (as you reported).  If we remove the listen socket from processing, then no one else can connect to this listen port any longer.  Seems like a form of DOS attack.  Of course the current behavior isn't acceptable either.  But in order to determine the best fix, I'd like to confirm what caused it, and ideally be able to reproduce it.

How do you know that problem started with a disconnect during TLS handshake?  Can you reproduce this?

Thank you for patience on this.

Scott

MatteoAns:
I believe there has been a misunderstanding: the issue occurred on the client side, not the server side. Currently, I haven't been able to reproduce it consistently except by forcing the state through an attached debugger.

Matteo

PR #462, opened 2026-02-27
Consider known parameters when comparing URIs

mykhailopopivniak:
Previously, URI comparison logic treated URIs with different known parameter values as equal.

Per RFC 3261 §19.1.4, parameters present in both URIs must match.

Add related unit tests.

Relates to #385.

### DIFF
--- a/resip/stack/Uri.cxx
+++ b/resip/stack/Uri.cxx
@@ -429,10 +429,17 @@ bool Uri::compareUriParametersEqual(Parameter* param1, Parameter* param2)
       case ParameterTypes::method:
       case ParameterTypes::maddr:
       case ParameterTypes::transport:
+      case ParameterTypes::rinstance:
+      case ParameterTypes::comp:
+      case ParameterTypes::extension:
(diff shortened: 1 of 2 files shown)


mykhailopopivniak:
`rinstance` seems to be analogous to `+sip.instance`, just for URI instead of Contact. I don't think there is point in comparing the whole URI if we can just check `rinstance` param and know that this contact is ours. I wouldn't remove `ClientRegistration::rinstanceIsMine`.

> DataParameter is actually capable of either being quoted or unquoted - sigcompid should probably be using it. That being said, due to the complications of the URN comparison, I think it's fine to leave it for now.

I suggest using the current DataParameter URI comparsion logic for it. 

RFC 5049
> As a result, the SIP/SigComp application SHOULD provide
> lexically equivalent URNs in each registration it generates. This is
> likely to be normal behavior in any case; applications are not likely
> to modify the value of their SIP/SigComp identifiers so that they
> remain functionally equivalent yet lexicographically different from
> previous identifiers.

At least we would catch obviously different URNs.

PR #470, opened 2026-06-17
Fix ACK correlation for retransmitted INVITE 2xx responses

portsip:
Summary:

InviteSession previously decided whether an inbound ACK should stop retransmission of an outstanding 200 OK to an INVITE or re-INVITE by comparing mLastRemoteSessionModification's CSeq against the ACK's CSeq.

The previous logic was effectively:

lastRemote.CSeq > ack.CSeq means drop the ACK as stale.

That is the wrong correlator. The response currently being retransmitted is held in mInvite200, and the ACK for a 2xx INVITE response carries the same CSeq sequence number as the INVITE and 2xx response it acknowledges.

This change correlates the ACK against mInvite200 instead of mLastRemoteSessionModification, applies the same key to the Retransmit200 and WaitForAck timers, and adds two defensive checks.

Symptoms:

A UAS-sent 200 OK to a re-INVITE may continue retransmitting even though the peer's ACK is received and routed to the dialog.

The log may show dropped stale ACK, or there may be no effective clear of the retransmission state.

onAckReceived is not called for the valid ACK.

After the ACK wait timer expires, the call may be torn down.

This was observed in production during session-timer re-INVITE glare and overlapping session-modification traffic.

Root cause:

The clear-on-ACK logic in several ACK handling paths was keyed on mLastRemoteSessionModification.

That value represents the most recently received remote session-modification request. It is not necessarily the INVITE whose 200 OK is currently being retransmitted.

During re-INVITE glare or overlapping session-modification traffic, the dialog state and mLastRemoteSessionModification can advance while a previous 200 OK is still waiting for its ACK. In that case, a valid ACK for the outstanding 200 OK can be misclassified as stale, leaving mCurrentRetransmit200 active until the ACK wait timer expires.

The correct correlator is the CSeq sequence number in mInvite200, because mInvite200 is the actual 2xx response currently being retransmitted.

Fix:

Add InviteSession::isAckForCurrent200(const SipMessage&) const.

The helper returns true only when all of the following are true:

1. The supplied message is a SIP request.
2. The request method is ACK.
3. A 2xx INVITE retransmission is currently outstanding.
4. The ACK's CSeq sequence number matches mInvite200's CSeq sequence number.

Replace the old mLastRemoteSessionModification-based stale ACK guard with !isAckForCurrent200(msg) in the relevant ACK handling paths.

Key startRetransmit200Timer() and the WaitForAck timer off mInvite200's CSeq instead of mLastRemoteSessionModification's CSeq.

Harden the Retransmit200 timer path so it retransmits only when all of the following are true:

mCurrentRetransmit200 is active.
mInvite200 is available.
timeout.seq() matches mInvite200->header(h_CSeq).sequence().

This prevents a stale timer event from retransmitting a replaced mInvite200.

Update the stale ACK log message to:

dropped stale or duplicate ACK

### DIFF
--- a/resip/dum/InviteSession.cxx
+++ b/resip/dum/InviteSession.cxx
@@ -1385,7 +1385,9 @@ InviteSession::dispatch(const DumTimeout& timeout)
 {
    if (timeout.type() == DumTimeout::Retransmit200)
    {
-      if (mCurrentRetransmit200)
+      if (mCurrentRetransmit200 &&
+          mInvite200 &&
+          timeout.seq() == mInvite200->header(h_CSeq).sequence())
       {
          InfoLog(<< "Retransmitting: " << endl << mInvite200->brief());
          //DumHelper::setOutgoingEncryptionLevel(*mInvite200, mCurrentEncryptionLevel);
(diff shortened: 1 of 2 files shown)


portsip:
This PR was created with the help of Claude Code and ChatGPT.

PR #476, opened 2026-07-16
Preserve percent-escaping in the URI user part

mykhailopopivniak:
### Problem

`Uri` decodes percent-escapes in the user part at parse time
(`dataUnescaped` into `mUser`) and re-encodes them on serialization via the
process-global user-encoding table. When that table is configured **not** to
escape a character through `setUriUserEncoding()` — e.g. to let a literal `#`
pass through unescaped — an inbound percent-escaped form of that character is
silently unquoted on output: `sip:1234%2300@h` parses `mUser` to `1234#00` and
then serializes as `sip:1234#00@h`. The user part no longer round-trips, and a
strict RFC 3261 parser rejects the result because `#` is not allowed unescaped
in the user part.

### Fix

Keep the as-received user-part bytes (`mUserRaw`) whenever the user part
contains a percent-escape, and emit them verbatim on encode as long as they
still decode (via the same `dataUnescaped`) to the current `mUser`. A
post-parse modification of the user part breaks that equality and falls back to
the existing table-based escaping. The raw bytes are propagated across the copy
constructor, `operator=`, and `getAorAsUri`, and cleared in
`setUserAsTelephoneSubscriber`.

An escaped character now round-trips unchanged regardless of the
encoding-table configuration, while an unescaped character is still encoded per
the table exactly as before. The common path is unaffected — `mUserRaw` is only
populated when the user part contains a `%`.

### Compatibility

This also changes serialization under the **default** encoding table: a user
part containing any percent-escape is now emitted with its escaping preserved
rather than canonicalized (e.g. `sip:12%2a34@h` stays `12%2a34` instead of
becoming `12*34`). This is RFC 3261-equivalent — user parts compare after
unescaping and `operator==`/`aorEqual` compare the decoded value — but it is a
visible on-wire change for every caller, not only those that customize the table
via `setUriUserEncoding()`.

### Tests

`testUri` extended: an escaped `#` survives a host retarget, a copy, an
assignment, and `getAorAsUri`; a post-parse modification falls back to
table-based escaping; a password-bearing user part and a lowercase escape are
preserved byte-for-byte.

### DIFF
--- a/resip/stack/Uri.cxx
+++ b/resip/stack/Uri.cxx
@@ -81,7 +81,8 @@ Uri::Uri(const Uri& rhs,
      mCanonicalHost(rhs.mCanonicalHost),
      mIsBetweenAngleQuotes(rhs.mIsBetweenAngleQuotes),
      mEmbeddedHeadersText(rhs.mEmbeddedHeadersText.get() ? new Data(*rhs.mEmbeddedHeadersText) : 0),
-     mEmbeddedHeaders(rhs.mEmbeddedHeaders.get() ? new SipMessage(*rhs.mEmbeddedHeaders) : 0)
+     mEmbeddedHeaders(rhs.mEmbeddedHeaders.get() ? new SipMessage(*rhs.mEmbeddedHeaders) : 0),
+     mUserRaw(rhs.mUserRaw ? new Data(*rhs.mUserRaw) : nullptr)
 {}
 
 
@@ -359,6 +360,14 @@ Uri::operator=(const Uri& rhs)
       mHostCanonicalized = rhs.mHostCanonicalized;
       mCanonicalHost = rhs.mCanonicalHost;
       mUser = rhs.mUser;
+      if (rhs.mUserRaw)
+      {
+         mUserRaw.reset(new Data(*rhs.mUserRaw));
+      }
+      else
+      {
+         mUserRaw.reset();
+      }
       mUserParameters = rhs.mUserParameters;
       mPort = rhs.mPort;
       mPassword = rhs.mPassword;
@@ -866,6 +875,14 @@ Uri::aorEqual(const resip::Uri& rhs) const
 void 
 Uri::getAorInternal(bool dropScheme, bool addPort, Data& aor) const
 {
+   // The AOR is an identity/lookup key, so the user part is intentionally
+   // emitted in canonical (table-escaped) form and is NOT preserved verbatim
+   // the way encodeParsed()/getAorAsUri() preserve the sender's mUserRaw.
+   // Canonicalizing here keeps the key stable regardless of how a sender
+   // escaped it (e.g. "%23" vs a literal '#'), which matters because
+   // getAorNoPort() feeds auth key comparisons (DigestAuthenticator, etc.)
+   // and the registration DB is keyed on this form. Do not change this to
+   // preserve mUserRaw without auditing those key/identity paths.
    checkParsed();
    // canonicalize host
 
@@ -1017,6 +1034,7 @@ Uri::setUserAsTelephoneSubscriber(const Token& telephoneSubscriber)
    mUser.clear();
    oDataStream str(mUser);
    str << telephoneSubscriber;
+   mUserRaw.reset();
 }
 
 Data
@@ -1041,8 +1059,14 @@ Uri::getAorAsUri(TransportType transportTypeToRemoveDefaultPort) const
    //.dcm. -- tel conversion?
    checkParsed();
    Uri ret;
-   ret.scheme() = mScheme;   
+   ret.scheme() = mScheme;
    ret.user() = mUser;
+   // Carry the as-received user bytes so a derived URI that reaches the wire
+   // (e.g. a Route built from the AOR) preserves the sender's escaping too.
+   if (mUserRaw)
+   {
+      ret.mUserRaw.reset(new Data(*mUserRaw));
+   }
    ret.host() = mHost;
 
    // Remove any default ports (if required)
@@ -1075,6 +1099,9 @@ Uri::parse(ParseBuffer& pb)
 {
    pb.skipWhitespace();
    const char* start = pb.position();
+
+   // Cleared up front so a path that sets no user part leaves no stale raw.
+   mUserRaw.reset();
 
    // Relative URLs (typically HTTP) start with a slash.  These
    // are seen when parsing the WebSocket handshake.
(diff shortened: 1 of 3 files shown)


sgodin:
I haven't had time to fully review this, but I did ask Claude AI about it and this was the response.  I figured I'd post it and let you consider it before I have time to fully review.   Note: it's point about getAOR is questionable, I need to think about whether we want getAOR() to preserve the escaping or stay canonical.

Verdict
The diagnosis is correct and the general approach is close to the only one that works given resiprocate's public API — but the fix is incomplete in one place, and it changes behavior under the default configuration in a way the PR description doesn't mention. I'd ask for changes rather than merge as-is.

The bug is real
The pipeline in Uri.cxx really is lossy. parse() calls pb.dataUnescaped(mUser, start), so %23 becomes a bare # in mUser, and encodeParsed() re-renders it through the process-global table from getUserEncodingTable(). If an application has called setUriUserEncoding('#', false) — which the existing test at testUri.cxx:1066 shows is a supported thing to do for Asterisk/Cisco interop — then an inbound %23 is silently unquoted to a bare # on output. That's a genuine round-trip failure, and it's not recoverable after the decode, because nothing in mUser distinguishes a received-escaped # from a received-literal #. Some form of "remember the as-received bytes" is required. Since user() hands out a mutable Data& (Uri.hxx:42), there's no setter to hook, so the "does the raw still decode to the current value?" check at encode time is a reasonable workaround rather than laziness.

The implementation is also careful in the ways that usually get missed: the raw capture spans exactly start→pb.position(), which is the same range dataUnescaped consumes (the colon when a password follows, else the @), so the invariant holds; the copy constructor, operator=, and setUserAsTelephoneSubscriber are all handled; member init order matches the declaration order, so no -Wreorder.

As a bonus the author doesn't claim: this also fixes an existing corruption under the default table. dataUnescaped deliberately leaves %3A, %7F, and control-char escapes undecoded (ParseBuffer.cxx), so sip:12%3A34@h currently leaves a literal % in mUser, which escapeToStream then re-escapes into %253A. Raw preservation makes that round-trip correctly.

What I'd want fixed before merge
getAorInternal() is not covered. Uri.cxx:906 does its own mUser.escapeToStream(str, getUserEncodingTable()) and the PR doesn't touch it. So after this change, with # encoding disabled, Data::from(uri) yields sip:1234%2300... while uri.getAOR() on the same object yields sip:1234#00.... The PR even propagates mUserRaw through getAorAsUri() while leaving getAOR() inconsistent with it. Whatever the right answer is for the AOR (canonical decoded form is defensible), the two paths shouldn't disagree, and that decision should be explicit.

It changes default-config output. Uri("sip:12%2a34@lvdx.com") currently encodes to sip:12*34@... (canonicalized, since * is safe in the table); with this PR it stays sip:12%2a34@.... The PR's own new test asserts exactly this. It's RFC-equivalent — 3261 §19.1.4 compares user parts after unescaping, and operator==/aorEqual still compare mUser, so matching is unaffected — but it's a visible on-wire change for every user of the library, not just the ones who touched the encoding table. That's a much wider blast radius than "preserve escaping when the table is customized," and it belongs in the PR description and probably in the release notes. It's inherent to the design, not a bug: you can't tell "escape that must be preserved" from "escape that's safe to canonicalize" without knowing the grammar, and the whole point of the knob is to emit characters the grammar forbids.

Cost on the encode path. Every encodeParsed() of a URI whose user part contains a % now builds a ParseBuffer, re-decodes, and compares. For short user parts Data's inline buffer avoids a heap allocation, so it's maybe ~100ns, and only for escaped users — but this is the message-serialization hot path in a proxy, and the check is repeated on every send. The %-scan loop the PR adds in parse() also duplicates a scan dataUnescaped already does internally. Worth a comment acknowledging the tradeoff at minimum.

ABI. It adds a private member to an exported class, so sizeof(Uri) changes. Fine for master, relevant if this is ever considered for a stable branch.

One minor thing: mUserRaw isn't reset on the no-user-part path of parse(). I couldn't construct a live path where that leaves stale state (parsing is once-only via checkParsed(), and the Uri(const Data&) route goes through operator=), so it's defensive rather than a bug — but a reset at the top of parse() would be cheaper to reason about than the case analysis I just did.

sgodin:
@mykhailopopivniak - Thanks for the quick turnaround — the revision addresses the 
three actionable points cleanly:

- **Default-config change documented.** The new *Compatibility* section spells
  out that `sip:12%2a34@h` now stays `12%2a34` instead of canonicalizing to
  `12*34`, for every caller and not only those who customize the table. That was
  my main concern about hidden behavior; it's now explicit.
- **Encode-path cost noted.** The comment in `encodeParsed` names the re-decode +
  compare, the inline-buffer avoidance of a heap alloc, and the unescaped-user
  fast path. No algorithmic change needed — just wanted the tradeoff on record.
- **Stale-raw guard.** `mUserRaw.reset()` up front in `parse()`, with the now-
  redundant `else { reset }` dropped from the capture block, is cleaner than the
  conditional I'd have reasoned through.

On the one item I left open — whether `getAOR()`/`getAorInternal()` should
preserve or canonicalize — I think the right call is to **keep it canonical and document the
divergence as intentional**, rather than make the two paths match. Rationale:

- Equality and map ordering (`operator==`, `operator<`) compare the *decoded*
  `mUser`, which this PR doesn't touch. So `mUserRaw` is invisible to every AOR
  *lookup/identity* path — the registration DB (`std::map<Uri,…>`) and the
  auth key comparisons behave identically before and after.
- `getAorNoPort()` feeds `DigestAuthenticator` (`getAorNoPort() == user`) and
  other auth key checks. Keeping `getAorInternal` canonical keeps that key byte-
  identical to today. Making it preserve escaping would change the auth key for
  any AOR containing an escaped character — an avoidable risk on a security path.
- `getAorAsUri()` preserving (for a `Route` that reaches the wire) while
  `getAOR()`/`getAorNoPort()` canonicalize (as a stable identity string) is a
  coherent split: on-wire fidelity vs. stable key.

Could you add a one-line comment on `getAorInternal` noting the canonicalization
is intentional and deliberately differs from `getAorAsUri`, so it isn't
"fixed" into preservation later? With that, I'm good to merge.

Here's a possible comment to add to the top of getAorInternal:

   // The AOR is an identity/lookup key, so the user part is intentionally
   // emitted in canonical (table-escaped) form and is NOT preserved verbatim
   // the way encodeParsed()/getAorAsUri() preserve the sender's mUserRaw.
   // Canonicalizing here keeps the key stable regardless of how a sender
   // escaped it (e.g. "%23" vs a literal '#'), which matters because
   // getAorNoPort() feeds auth key comparisons (DigestAuthenticator, etc.)
   // and the registration DB is keyed on this form. Do not change this to
   // preserve mUserRaw without auditing those key/identity paths.


Thanks for the contribution!

PR #477, opened 2026-07-29
Add urn: (RFC 8141) scheme support to resip::Uri

jmariasc:
**Motivation**

The urn: scheme (RFC 8141 ) is commonly used in IMS to request emergency services, where the UE requests for a specific service such as `urn:service:sos` (RFC 5031) and the network routes the call to the most appropriate PSAP. The UAC must place the URN in the Request-URI, and optionally in the To header, and all the B2BUAs and SIP proxies involved in the call path must be capable of handling it.

Since resiprocate/resip could not previously process SIP messages containing a URN in the URI because of resip::Uri had no notion of the urn: scheme, and DUM rejected such request-URIs, I have added some basic support for this feature for your review and consideration.

**Details**

- Add urn: (RFC 8141) scheme support to resip::Uri and registers urn in Symbols; 
- Add it's own ABNF EncodingTable. Parses/encodes it as an NID:NSS blob with its own ABNF, distinct from sip:/sips:/tel:; 
- operator== compares NID case-insensitively and NSS case-sensitively per RFC 8141 instead of a whole-string compare; 
- Reuses mUserRaw to preserve as-received escaping so literal vs. escaped #/? don't collapse on re-encode; 
- Adds unit tests for parsing, encoding and equality.
- Added the urn scheme to the MasterProfile in DUM and registers urn as a supported scheme so DialogUsageManager::validateRequestURI() no longer rejects an INVITE targeting a urn: URI with 416 Unsupported URI Scheme.
- Sanity checks before copying the user part from a urn to a sip Uri because of BaseCreator/Dialog assumed sip-like user-part semantics; since urn's ABNF differs, guards prevent copying an invalid user part when deriving a sip Uri from a urn Uri.

### DIFF
--- a/resip/dum/Dialog.cxx
+++ b/resip/dum/Dialog.cxx
@@ -125,20 +125,27 @@ Dialog::Dialog(DialogUsageManager& dum, const SipMessage& msg, DialogSet& ds)
                      {
                         // Contact header will always have a sip: scheme so it's safe to copy
                         // the user part from sip:, sips: and tel: but not from others like urn:
-                        if (isEqualNoCase(request.header(h_To).uri().scheme(), Symbols::Sip) ||
-                           isEqualNoCase(request.header(h_To).uri().scheme(), Symbols::Sips) ||
-                           isEqualNoCase(request.header(h_To).uri().scheme(), Symbols::Tel))
+                        // (RFC 8141's NID:NSS has nothing to do with a sip: user's ABNF).
+                        // When skipped, mLocalContact.uri().user() keeps whatever
+                        // getOverrideHostAndPort() set above (or stays empty) --
+                        // e.g. for a urn: To (an incoming request targeting an
+                        // emergency-services URN), the local Contact ends up with
+                        // no user part at all: "sip:host:port". Legal per RFC 3261
+                        // (Contact only needs a valid sip:/sips: URI), and arguably
+                        // the only sane choice here, since there is no sip:-shaped
+                        // user to carry over from a urn: To.
+                        if (request.header(h_To).uri().isUserRelevant())
                         {
                            mLocalContact.uri().user() = request.header(h_To).uri().user();
                         }
                      }
                      else
                      {
-                        // Contact header will always have a sip: scheme so it's safe to copy
-                        // the user part from sip:, sips: and tel: but not from others like urn:
-                        if (isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Sip) ||
-                           isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Sips) ||
-                           isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Tel))
+                        // Same reasoning as above, for the case where the
+                        // Request-URI (rather than To) carries the user to copy
+                        // -- e.g. a urn: Request-URI (RFC 5031 emergency call)
+                        // also leaves mLocalContact.uri().user() untouched.
+                        if (request.header(h_RequestLine).uri().isUserRelevant())
                         {
                            mLocalContact.uri().user() = request.header(h_RequestLine).uri().user();
                         }
--- a/resip/dum/Dialog.cxx
+++ b/resip/dum/Dialog.cxx
@@ -123,11 +123,25 @@ Dialog::Dialog(DialogUsageManager& dum, const SipMessage& msg, DialogSet& ds)
                      }
                      if(request.header(h_RequestLine).uri().user().empty())
                      {
-                        mLocalContact.uri().user() = request.header(h_To).uri().user(); 
+                        // Contact header will always have a sip: scheme so it's safe to copy
+                        // the user part from sip:, sips: and tel: but not from others like urn:
+                        if (isEqualNoCase(request.header(h_To).uri().scheme(), Symbols::Sip) ||
+                           isEqualNoCase(request.header(h_To).uri().scheme(), Symbols::Sips) ||
+                           isEqualNoCase(request.header(h_To).uri().scheme(), Symbols::Tel))
+                        {
+                           mLocalContact.uri().user() = request.header(h_To).uri().user();
+                        }
                      }
                      else
                      {
-                        mLocalContact.uri().user() = request.header(h_RequestLine).uri().user(); 
+                        // Contact header will always have a sip: scheme so it's safe to copy
+                        // the user part from sip:, sips: and tel: but not from others like urn:
+                        if (isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Sip) ||
+                           isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Sips) ||
+                           isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Tel))
+                        {
+                           mLocalContact.uri().user() = request.header(h_RequestLine).uri().user();
+                        }
                      }
                      const Data& instanceId = mDialogSet.mUserProfile->getInstanceId();
                      if (!contact.uri().exists(p_gr) && !instanceId.empty())
--- a/resip/dum/MasterProfile.cxx
+++ b/resip/dum/MasterProfile.cxx
@@ -38,6 +38,7 @@ MasterProfile::MasterProfile() :
    addSupportedMethod(BYE);
    addSupportedMethod(UPDATE);
    addSupportedScheme(Symbols::Sip);  
+   addSupportedScheme(Symbols::Urn);
 }
 
 void 
--- a/resip/dum/MasterProfile.cxx
+++ b/resip/dum/MasterProfile.cxx
@@ -37,8 +37,10 @@ MasterProfile::MasterProfile() :
    addSupportedMethod(OPTIONS);
    addSupportedMethod(BYE);
    addSupportedMethod(UPDATE);
-   addSupportedScheme(Symbols::Sip);  
-   addSupportedScheme(Symbols::Urn);
+   addSupportedScheme(Symbols::Sip);
+   // urn: (e.g. RFC 5031 emergency service URNs) is intentionally not
+   // added to the default supported schemes. An application that wants 
+   // it should call addSupportedScheme(Symbols::Urn) itself.
 }
 
 void 
--- a/resip/dum/BaseCreator.cxx
+++ b/resip/dum/BaseCreator.cxx
@@ -105,7 +105,14 @@ BaseCreator::makeInitialRequest(const NameAddr& target, const NameAddr& from, Me
       {
          contact.uri() = mUserProfile->getOverrideHostAndPort();
       }
-      contact.uri().user() = from.uri().user();
+      // Contact header will always have a sip: scheme so it's safe to copy
+      // the user part from sip:, sips: and tel: but not from others like urn:
+      if (isEqualNoCase(from.uri().scheme(), Symbols::Sip) ||
+         isEqualNoCase(from.uri().scheme(), Symbols::Sips) ||
+         isEqualNoCase(from.uri().scheme(), Symbols::Tel))
+      {
+         contact.uri().user() = from.uri().user();
+      }
 
       // .jjg. there isn't anything in the outbound [11] draft that says we 
       // aren't allowed to include p_Instance in this case...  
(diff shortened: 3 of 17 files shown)


resip/dum/MasterProfile.cxx
```diff
@@ -38,6 +38,7 @@ MasterProfile::MasterProfile() :
    addSupportedMethod(BYE);
    addSupportedMethod(UPDATE);
    addSupportedScheme(Symbols::Sip);  
+   addSupportedScheme(Symbols::Urn);
```
sgodin: This is a behavior change that most likely requires support in the application to be fully handled properly.  I thinking we should remove Urn from the default MasterProfile and require applications to add it specifically if they want to support URN schemes.

resip/dum/Dialog.cxx
```diff
@@ -123,11 +123,25 @@ Dialog::Dialog(DialogUsageManager& dum, const SipMessage& msg, DialogSet& ds)
...
+                           mLocalContact.uri().user() = request.header(h_To).uri().user();
+                        }
                      }
                      else
                      {
-                        mLocalContact.uri().user() = request.header(h_RequestLine).uri().user(); 
+                        // Contact header will always have a sip: scheme so it's safe to copy
+                        // the user part from sip:, sips: and tel: but not from others like urn:
+                        if (isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Sip) ||
+                           isEqualNoCase(request.header(h_RequestLine).uri().scheme(), Symbols::Sips) ||
```
sgodin: Since checking these 3 scheme's is done in 3 places - it would make sense to add a helper method to URI instead.  Something like: bool Uri::isUserRelevant().

sgodin:
@jmariasc  Thank you for this contribution!  URN support is a nice addition to the stack.  

I asked Claude to analyze this PR and this was the response...

Summary: the approach is sound and unusually well-commented, but there are a few real problems.

Blocking
1. A bare (unbracketed) urn: in a header swallows the header parameters — Uri.cxx:1176-1190

In the mIsBetweenAngleQuotes == false branch the parser stops only at whitespace. NameAddr::parse() calls mUri.parse(pb) and then parseParameters(pb) on whatever is left (NameAddr.cxx:250), so:

To: urn:service:sos;tag=abc123
parses as user() == "service:sos;tag=abc123" with no tag parameter → To-tag lost → dialog matching breaks.

The in-code justification ("SIP requires angle brackets exactly to avoid ';' clashing with NameAddr params") is only half true: RFC 3261 §20 requires brackets when the URI contains ,, ? or ;. urn:service:sos contains none of those, so the bare form is legal on the wire — and it's exactly the URN this PR targets. The PR's own test only covers <urn:service:sos>;tag=abc123, so it passes.

Suggested fix: in the non-bracketed branch also stop at ;,? (a URN needing those chars must be bracketed anyway, where the permissive rule still applies). Same issue in Aor.cxx:34. Please add NameAddr na("urn:service:sos;tag=abc123") to the tests.

2. Aor's urn branch leaves mPort/mOldPort uninitialized — Aor.cxx:29-37

Aor::Aor(const Data&) has no member-init list and mPort has no in-class initializer; the new branch returns before mPort = 0. Aor::value() then reads mOldPort != mPort on uninitialized ints. This is pre-existing in the tel: branch right below, but the new code copies it. One-line fix: set mPort = 0; before the return (or fix the constructor properly).

3. Case-insensitive NID equality is inconsistent with <, the hash, and the AOR — Uri.cxx:530-541

operator== now says urn:EXAMPLE:abc == urn:example:abc, but:

Uri::operator< compares mUser case-sensitively → std::set/std::map<Uri> will hold both copies.
HashValueImp(resip::Uri, Data::from(data).hash()) is case-sensitive → equal objects, different hashes; HashMap<Uri,…> is broken.
getAorInternal() doesn't lowercase the NID, so registration/auth keys and Aor::operator== (which compares the value() string) stay case-sensitive.
Cleanest fix: lowercase the NID at parse time, the same way schemeLowercase() already normalizes the scheme. RFC 8141 makes NIDs case-insensitive so that's spec-legal, and it makes ==, <, hash and AOR all agree without the special-case comparator.

Design question for the maintainers
4. addSupportedScheme(Symbols::Urn) in the MasterProfile constructor — MasterProfile.cxx:41

The default list today is just sip (not even sips). This makes every DUM application stop returning 416 for URN Request-URIs and start feeding them into Dialog/DialogSet. Whether to accept emergency URNs is deployment policy; I'd rather the emergency-services app call addSupportedScheme(Symbols::Urn) itself than change the library default. Same argument, more weakly, for the "empty list means sip/sips/tel" default in MessageFilterRule.cxx:97. (ServerRegistration is unaffected — it still hard-requires sip/sips.)

5. No URN validation at all. Anything after urn: up to whitespace/> is accepted: urn: (empty), urn:foo (no NSS), urn:-bad-:x, a >32-char NID, urn:urn:x (forbidden by RFC 8141 §5.1). Combined with number 4, malformed URNs now reach the application instead of getting a 400/416. At minimum require a non-empty alphanumeric-leading NID followed by :, and throw ParseException otherwise — which also removes the need for urnUserEqual's "no colon found" fallback.

Smaller items

- RFC 8141 §3.1 excludes r-/q-/f-components from URN equivalence, but they're folded into mUser, so urn:example:a?=b != urn:example:a. Probably fine for SIP routing, but the code cites the RFC for the case rules and should note this deviation.
- encodeParsed()'s urn branch (Uri.cxx:1426-1455) is a verbatim copy of the generic mUserRaw branch immediately below it; the only difference is the fallback table. Collapse to const EncodingTable& t = (mScheme == Symbols::Urn) ? getUrnEncodingTable() : getUserEncodingTable(); and reuse the existing block — ~30 lines and one divergence risk gone. The %-scan/mUserRaw block in parse() duplicates Uri.cxx:1184-1206 similarly.
- urnUserEqual allocates 4 Data copies (substr ×4) plus a temp Data for Symbols::COLON in find(), on the dialog-matching hot path. Add an early if (a.size() != b.size()) return false; and compare in place.
- The Sip || Sips || Tel triplet is copy-pasted three times (BaseCreator once, Dialog twice). A helper on Uri would read better and keeps the list in one place.
- Case-sensitivity nit: getAorInternal() and encodeParsed() test mScheme == Symbols::Urn case-sensitively while operator== uses isEqualNoCase. Only matters for hand-built URIs (uri.scheme() = "URN"), which the tests never exercise. Pick one.
- With a URN Request-URI, Dialog now leaves the local Contact with no user part (sip:host:port). Legal and probably intended, but worth a comment/test. Note it also changes behavior for getOverrideHostAndPort(): its user part is now preserved rather than overwritten for non-sip/sips/tel From URIs.

The test additions are good quality and testUri is wired into CMake/runtests.sh, so they will run. Missing coverage: the bare-URN-with-params case (number 1, currently fails), malformed URNs, Aor("urn:…"), and </hash consistency.

resip/dum/MasterProfile.cxx
jmariasc: I initially thought the same having some doubts about it. In fact, it was in a later commit when I decided to  introduce it. I thought this way it was not necessary to touch repro for example. But you are right, it would be better to explore before how repro could be affected when trying to handle a urn instead assuming it'll be transparent.

resip/dum/Dialog.cxx
jmariasc: Sure, it makes sense.

jmariasc:
@sgodin Thank you for reviewing this. Only to say I have already implemented the urn support (several approaches) in personal branches since 2020 with success. Let me digest all this stuff and I'll come back to you. I'll ask Claude for some help by sure too ;-)

jmariasc:
@sgodin after some rounds, I think all the issues were addressed.  Please find the result. I asked Claude for some help to write it.

1. Bare urn: swallowing header params. Confirmed and fixed, but your suggested one-liner ("also stop at ;,? in the non-bracketed branch") would have introduced a new regression: RequestLine::parse() calls Uri::parse() with the exact same mIsBetweenAngleQuotes == false state as NameAddr::parse()'s bare branch, and a Request-URI can never be bracketed — so blanket-stopping at ;/?/, there would truncate a legitimate urn:example:a;b;c Request-URI. A second flag was added, Uri::setIsBareAddrSpec(), set only by NameAddr::parse()'s unbracketed branch, so the Request-Line/standalone-construction/bracketed cases keep the permissive (whitespace-only) rule while the true ambiguous case (bare addr-spec inside a header) now stops at ;/?/,. Added NameAddr na("urn:service:sos;tag=abc123") plus a Request-Line regression test with a literal ; in the NSS to guard both directions.

Deliberately did not apply the same change to Aor.cxx:34 — Aor never has trailing header parameters, so there's no ambiguity to resolve there; doing it anyway would just reintroduce the same truncation bug for no benefit.

2. Aor::mPort/mOldPort uninitialized. Fixed with in-class default initializers (Aor.hxx). While wiring up testAor.cxx to actually verify this (it turned out to never have been registered in CMake at all — it wasn't compiling or running before this), I hit a second, unrelated pre-existing bug: Aor::Aor(const Data&) never copies its input — mScheme/mUser/mHost are non-owning views (ParseBuffer::data()) into whatever buffer value wraps. Aor aor("sip:...") with a literal creates a temporary Data that's destroyed right after the constructor returns, leaving dangling pointers. Fixed by adding Aor::mOwnedBuffer (owns a copy, same lifetime as the object) and parsing over that instead of value directly. Verified with valgrind (0 errors) and a standalone repro with a heap-allocated Data destroyed immediately after construction.

3. NID case-insensitivity inconsistent with <, hash, AOR. Went with your suggested approach: the NID is now lowercased at parse time (mirrors schemeLowercase()), and urnUserEqual was removed entirely — operator==, operator<, std::hash<Uri>, and getAorInternal() all just read the now-normalized mUser, so they agree by construction. Verified with urn:EXAMPLE:abc vs urn:example:abc: equal, same ordering, same hash, same getAor(), collapse to one entry in std::set<Uri>.

4. Reverted addSupportedScheme(Symbols::Urn) from MasterProfile's constructor. Documented the opt-in with a Doxygen comment + code example on MasterProfile::addSupportedScheme() itself, since that's the API surface an app author would actually be looking at.

5. Added NID syntax validation in Uri::parse()'s urn branch, throwing ParseException (via pb.fail(), same idiom the rest of the file uses) for: missing NID:NSS separator, NID outside the 2–32 char range, invalid NID characters, - at the start/end of NID, empty NSS, and NID == "urn" (case-insensitive). This did let us drop the "no colon found" fallback you flagged in urnUserEqual — moot now since it's been removed anyway.

6. Documented the RFC 8141 §3.1 deviation (r-/q-/f-components folded into the equivalence) as a deliberate comment next to the user-part comparison in operator==, rather than changing behavior.

7. Collapsed encodeParsed()'s urn branch into the generic one — a single table reference chosen by scheme, reusing one code path instead of two near-identical ones.

8. Extracted the %-scan/mUserRaw block from parse() into a shared decodeUserPreservingEscapes() helper, used by both the sip:/sips: and urn: branches.

9. Moot — urnUserEqual no longer exists (see #3).

10. Added Uri::isUserRelevant() and replaced all three copies of the Sip||Sips||Tel triplet (BaseCreator.cxx once, Dialog.cxx twice) with it.

11. Fixed — getAorInternal()/encodeParsed() now use isEqualNoCase(mScheme, Symbols::Urn) consistently with the rest of the file.

12. Added explanatory comments in both BaseCreator.cxx and Dialog.cxx on the empty-user-part Contact and the getOverrideHostAndPort() interaction. Also added an end-to-end DUM test (resip/dum/test/testUrnContact.cxx): a real UAC/UAS pair over UDP loopback, UAC sends an INVITE to urn:service:sos, and we assert the UAS's resulting Contact (as seen in the 180) is sip:host:port with no user part.

13. The NID character check used isalnum(), which is locale-dependent; under a non-C locale (not unusual in embedded/softswitch deployments that call setlocale()), it can accept 8-bit characters RFC 8141 doesn't allow. Replaced with an explicit ASCII range check.

14. testAor.cxx existed but was never registered in resip/stack/test/CMakeLists.txt — it had never actually been compiled or run. Wired it in.

There is a pre-existing issue in the code I'am flagging but not fixing here:

std::hash<Uri> can differ between two Uris that compare equal whenever the user part contains a %XX escape (it hashes the encoded form via encodeParsed(), which re-emits mUserRaw verbatim, rather than the canonical decoded mUser). I reproduced the identical pattern for sip: (sip:bob@example.com vs sip:b%6fb@example.com: equal, different hash) — it's a pre-existing design gap in Uri, not something urn: introduces; it just inherits it by reusing the same mUserRaw mechanism. Flagging it since it's a real violation of the hash/equality contract, but fixing it is a separate, broader change than this PR's scope.

Please review the changes and thank you for your time.

sgodin:
@jmariasc Thank you for the updates - looks good to me!

PR #480, opened 2026-08-10
Complete the CSPRNG migration for the remaining security relevant draws

Nexory:
## Summary

`Random::getRandom()` is a plain `random()` stream. With `RESIP_RANDOM_THREAD_LOCAL` and `RESIP_RANDOM_THREAD_MUTEX` both off by default there is a single process wide stream on POSIX, and `MultipartMixedContents::setBoundary()` puts eight raw bytes of it on the wire in every multipart body. Anything else drawn from that stream is therefore observable, not merely weak in the abstract.

The migration to `getCryptoRandom()` has run in three waves (#432, #433, then the STUN transaction ID and TURN channel number work). Three draws were left behind:

- `reflow/FlowManager.cxx` sets the X509 serial of the DTLS-SRTP certificate with `getRandom()`. It is the same line, the same comment and the same following assert as `BaseSecurity::generateUserCert()`, which #433 already moved over, and as `makeSelfCert()`. This was the third copy.
- `stunRandomPort()` still calls `stunRand()`, which `stunBuildReqSimple()` describes in its own comment as the weak generator.
- `setBoundary()` is the draw that publishes the stream.

I am not claiming any of these is exploitable on its own. The self signed DTLS certificate is pinned by its SDP fingerprint, so a predictable serial does not break it. The argument is consistency: the project has already decided twice that this exact line should use the CSPRNG.

On the cost raised in #431, I measured it instead of arguing it. `getCryptoRandomHex(8)` is about 28x slower than `getRandomHex(8)` here, roughly 580 ns against 20 ns, median of nine interleaved rounds of 200000 calls each. At 10000 multipart bodies per second that is 0.56 percent of one core. A certificate serial is drawn once, right next to an RSA key generation.

## Verification

`rutil/test/testRandomStreamIsolation` pins the stream with `srandom()` and checks three things, two of them controls so that a pass actually means something:

1. control: a wire visible draw repeats under a pinned seed, so the stream really is observable. Without this the rest would prove nothing.
2. control: `getCryptoRandomHex()` does not repeat, so `srandom()` has no hold on the CSPRNG.
3. under test: `stunRandomPort()` must not repeat.

Before the change the test fails with `stunRandomPort() returned 28704 twice under a pinned seed`, exit code 255. After the change it passes, exit code 0. It is skipped on Windows, where `getRandom()` already goes through `RtlGenRandom`.

`getCryptoRandom()` returns a full width int and can be negative, where the old path never was. `stunRandomPort()` masks down to 15 bits so the documented range still holds, and the test verifies that over 100000 draws, observing exactly 0x4000 to 0x7fff.

Full suite: 69 of 69 pass, before and after the change. `testMultipartMixedContents` and `testMultipartRelated` pass. `reflow` was configured and built separately so that the `FlowManager.cxx` change is actually compiled rather than merely asserted.

After the stun change nothing in the tree calls `stunRand()` any more, so the last commit marks it with the project's existing `RESIP_DEPRECATED` macro instead of removing it, since `Stun.hxx` is an installed public header. The intent is to keep this fix from being quietly reopened by a future caller. I checked that the attribute actually fires rather than sitting there silently: a deliberate call warns that `resip::stunRand()` is deprecated, the same call with the marker taken back out warns not at all, and the tree itself builds with zero deprecation warnings. Happy to drop that commit if you would rather leave the API untouched.

## Limitations

- `FlowManager::createCert()` is private, so the test cannot call it directly. That change is covered by compilation and by the identical, already migrated sibling in `BaseSecurity::generateUserCert()`, not by a test case of its own.
- `testSMIME` is not registered with ctest in the configuration I used, so it did not run here.
- I deliberately left the draws that are not security relevant alone: Retry-After jitter, timer jitter, DNS SRV weighting, presence tuple IDs and the rend load test code. Moving those would add cost for nothing.
- I left `stunRand()` itself in place because it is declared in the public header.

Happy to split this differently, drop the boundary commit, or put the CSPRNG draws behind a build flag if you would rather keep the fast path available.


sgodin:
@Nexory 

Thanks for the careful writeup — the performance numbers and the control-based
test design are both appreciated. I'd like to push back on the value of these
three changes, though.

The premise (raw `random()` output on the wire lets an observer reconstruct the
stream) is technically valid, but I think the payoff is nil at this point,
because the earlier PRs in this series already moved everything worth predicting
off that stream. `Helper::computeUniqueBranch()`, `computeCallId()` and
`computeTag()` are all on `getCryptoRandomHex` today (Helper.cxx:679-716). What
still draws from `random()` is `Retry-After` values, timer jitter, DNS SRV weight
selection, presence tuple IDs, and an ichat-gw SSRC. Predicting any of those
gains an attacker nothing, so disclosing the stream discloses nothing of
consequence. Note also that on Windows `Random::getRandom()` already goes through
RtlGenRandom, so this concern is POSIX-only to begin with.

Taking the three individually:

- **`stunRandomPort()`** — this is unreachable code. Its only callers are
  `stunTest`, `stunNatType` and `stunOpenSocketPair` within Stun.cxx, and none of
  those are called anywhere in the tree; the in-tree consumers of Stun.hxx
  (UdpTransport, repro's BasicWsConnectionValidator and CookieAuthenticator, tfm's
  StunServer) use only the parse/encode functions. Even if it were live, the local
  port ends up announced in SDP/Via anyway.

- **`setBoundary()`** — the boundary only has to not collide with part content,
  which 8 random bytes gives us from any generator. Repetition across messages is
  harmless.

- **`createCert()` serial** — random serials matter for CA-issued certs, where a
  predictable serial enables a chosen-prefix collision against the signature.
  These are self-signed ephemeral DTLS certs verified by SDP fingerprint (RFC
  8122), so there's no signing oracle and the peer hashes the whole cert. This one
  also has a downside: `getCryptoRandom()` returns a full-width `int`, so it can
  now produce negative serials, where `getRandom()` was always positive. RFC 5280
  §4.1.2.2 requires a positive serial. (Security.cxx:1562 and makeCert.cxx:98 have
  the same problem already — worth a separate fix masking all three with
  `& 0x7FFFFFFF`.)

The cost side concerns me more than the value side. `testRandomStreamIsolation`
hard-fails in two supported configurations — `-DWITH_SSL=OFF`, where
`getCryptoRandom()` falls through to `getRandom()` (Random.cxx:349, 481) so
controls 2 and 3 both fire, and with `RESIP_RANDOM_THREAD_LOCAL` /
`RESIP_RANDOM_THREAD_MUTEX` enabled (Random.hxx:25/32), where `srandom()` no
longer pins the stream and control 1 fires. It's also flaky: `stunRandomPort()`
has only 16384 possible values, so comparing two independent draws collides about
1 run in 16k, and the failure message would blame the seeded stream. Neither would
show up in CI, which only builds with SSL and skips the test on Windows.

Marking `stunRand()` deprecated is also a public API change, and since Windows
builds with `/sdl` (which elevates C4996 to an error), for MSVC consumers it's a
hard build break rather than a warning — for a function whose only in-tree caller
this PR itself just removed.

I'm inclined to close this rather than merge it.

Nexory:
You are right on every point. I checked each one against the tree rather than
taking it on trust, and my own numbers agree with yours.

- `stunRandomPort()` is dead code. `stunTest`, `stunNatType` and
  `stunOpenSocketPair` have no callers outside `Stun.cxx` at all. I checked the
  callers of `stunRandomPort` and stopped there instead of going one level up,
  which is exactly the check that mattered.
- The test is flaky, and worse than "about 1 in 16k": 135 collisions in 2,000,000
  paired draws, so 1 in 14,814.
- With `-DWITH_SSL=OFF` it does fail. Compiling `Random.cxx` without `USE_SSL`
  and pinning the seed, `getCryptoRandomHex(8)` returns the same value twice, so
  control 2 fires.
- With `RESIP_RANDOM_THREAD_LOCAL` it fails too. Rebuilt with the macro,
  `srandom()` no longer pins the stream and control 1 fires.
- The deprecation marker is a build break, not a warning. `add_compile_options(/W3 /sdl)`
  at CMakeLists.txt:1008 elevates C4996, so MSVC consumers would fail to compile
  a function whose last in-tree caller this PR removed.

On value I have nothing to argue. My own text said the DTLS certificate is
pinned by its SDP fingerprint and that I was not claiming exploitability; you
drew the conclusion I should have drawn from that.

One correction in your favour. The negative serial is not a possibility, it is
the normal case: `getCryptoRandom()` fills a whole `int` from `RAND_bytes`, and
over 100,000 draws 49,851 were negative. Generating 40 certificates through
`generateUserCert` gave 15 with a serial that `BN_is_negative` reports as
negative, and OpenSSL prints them as such. So `Security.cxx:1562` and
`makeCert.cxx:98` have been emitting non-conforming serials since #433, and this
PR would have added a third site.

I have that fix ready as you described it, masking both existing sites with
`& 0x7FFFFFFF`, with a test that goes from 15 of 40 negative to 0 of 40 and
carries its own control so the zero means something. Happy to open it as a
separate PR whenever suits you.

Leaving this one open for you to close as you see fit, since the decision is
yours. Thanks for taking the time to go through it in this much detail.

sgodin:
@Nexory I agree the negative serial fix is needed.  I will close this PR and if you can open a new one with those fixes it would be greatly appreciated.  Thanks!

PR #481, opened 2026-08-10
Keep certificate serial numbers positive (RFC 5280 4.1.2.2)

Nexory:
Follow-up to #480, with just the part you asked for.

## What is wrong

RFC 5280 section 4.1.2.2 requires the certificate serial number to be a
positive integer. `Random::getCryptoRandom()` fills a whole `int` from
`RAND_bytes`, so it is negative about half the time, where the `getRandom()`
it replaced in #433 never was.

Measured here: 49851 of 100000 draws are negative. Generating 40 certificates
through `generateUserCert` produced 29 whose serial `BN_is_negative` reports
as negative, and `openssl x509 -noout -text` prints them as
`Serial Number: -12345 (-0x3039)`.

Two sites draw the serial from `getCryptoRandom()`:

- `resip/stack/ssl/Security.cxx:1562` in `generateUserCert()`. This is the
  live one, reached from repro through `CertSubscriptionHandler`.
- `resip/stack/makeCert.cxx:98`. Same line, but the file appears in no
  `CMakeLists.txt` and carries its own `main()`, so this one is for
  consistency rather than impact. I mention it because you named it, and
  because leaving one of two identical lines behind is how this came about.

`reflow/FlowManager.cxx` and `resip/stack/test/makeSelfCert.cxx` also set a
serial, but both still use `getRandom()` and are therefore always positive.
Nothing to do there.

## Verification

From a clean checkout of master, Debug build with assertions on:

- before: `testCertSerial` reports 29 of 40 certificates with a negative
  serial and exits 255
- after: 0 of 40, exits 0
- suite: 69 of 69, in Debug and in RelWithDebInfo

The test carries its own control, which builds a certificate with a negative
serial from scratch and confirms the probe sees it. That control matters: my
first version parsed a real certificate and wrote a negative serial into it
afterwards, which `i2d_X509()` ignores because it re-emits the DER cached
during parsing. It passed while the fix was absent, for the wrong reason,
and failed once the fix was in.

I took the three things that went wrong last time seriously:

- Determinism: 25 consecutive runs, no failures. The mask makes the outcome
  independent of the draw, so there is no collision window.
- `-DWITH_SSL=OFF`: the test is registered inside the existing
  `if(OPENSSL_FOUND)` block next to `testSecurity`, so it is simply absent
  there. I could not run that configuration to completion, because an
  unmodified master already fails it with
  `SipStack.cxx:450:29: error: expected type-specifier before 'DtlsTransport'`.
  That is pre-existing and unrelated, but it means my check there is limited
  to "the test is not registered" rather than "the suite is green".
- MSVC `/sdl`: the build produces no deprecation warnings. The test uses the
  same OpenSSL calls `Security.cxx` already uses, and the project's
  `OPENSSL_API_COMPAT=0x10101000L` covers them.

Thanks for the detailed review on the previous one; it was a much better use
of your time than a merge would have been.

### DIFF
--- a/resip/stack/makeCert.cxx
+++ b/resip/stack/makeCert.cxx
@@ -95,7 +95,9 @@ int makeSelfCert(X509 **cert, EVP_PKEY *privkey)   // should include a Uri type 
 
   //  RAND_bytes((char *) serial , 4);
   //serial = 1;
-  serial = Random::getCryptoRandom();  // get an int worth of randomness
+  // RFC 5280 4.1.2.2 wants a positive serial, and getCryptoRandom() fills a
+  // whole int from RAND_bytes, so it is negative about half the time.
+  serial = Random::getCryptoRandom() & 0x7FFFFFFF;
   ASN1_INTEGER_set(X509_get_serialNumber(selfcert),serial);
 
   X509_NAME_add_entry_by_txt( subject, "O",  MBSTRING_UTF8, (unsigned char *) domain.data(), domain.size(), -1, 0);
--- a/resip/stack/ssl/Security.cxx
+++ b/resip/stack/ssl/Security.cxx
@@ -1559,7 +1559,9 @@ BaseSecurity::generateUserCert (const Data& pAor, int expireDays, int keyLen )
    // set version to X509v3 (starts from 0)
    X509_set_version(cert, 2L);
    
-   int serial = Random::getCryptoRandom();  // get an int worth of randomness
+   // RFC 5280 4.1.2.2 wants a positive serial, and getCryptoRandom() fills a
+   // whole int from RAND_bytes, so it is negative about half the time.
+   int serial = Random::getCryptoRandom() & 0x7FFFFFFF;
    resip_assert(sizeof(int)==4);
    ASN1_INTEGER_set(X509_get_serialNumber(cert),serial);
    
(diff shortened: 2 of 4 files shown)


sgodin:
@Nexory - thanks for this.  There are build errors in the new unit tests:
https://github.com/resiprocate/resiprocate/actions/runs/31437502507/job/93837555120?pr=481

Also the OpenSSL code in the test harness should be guarded with USE_SSL preprocessor defines.  See other code in the repo that uses OpenSSL.

Nexory:
Fixed in 4703abe. Two things:

- Added `#include <openssl/rsa.h>` and `#include <openssl/evp.h>`. The RSA
  symbols were only reaching the file transitively via `openssl/x509.h`, which
  does not hold under the CI's OpenSSL headers (1.1.1 with
  `OPENSSL_API_COMPAT=0x10101000L` leaves `RSA_new` and friends undeclared).
  This matches the include list `ssl/Security.cxx` already uses.
- Guarded the whole body with `USE_SSL`, with a no-op `main` otherwise, like
  the other OpenSSL-using tests in the stack.

I reproduced the exact failure on `debian:bullseye-slim` (the CI image, OpenSSL
1.1.1w) and confirmed the target builds and the test passes there after the
change; the full stack suite stays at 69/69 on OpenSSL 3.5. My earlier local
build only had OpenSSL 3.5, whose headers pull rsa.h transitively, which is why
I missed it. Sorry for the round trip.

PR #482, opened 2026-08-14
Verify multipart/signed against the received bytes (RFC 1847)

Nexory:
## Summary

`BaseSecurity::checkSignature()` verified the PKCS#7 signature against a
re-encoding of the parsed first body part rather than against the bytes that
arrived. `Contents::encodeHeaders()` writes a fixed set of headers in a fixed
order with one space after the colon and rebuilds the Content-Type from the
parsed `Mime`, so a signature made by any peer whose header block differs in a
single byte fails to verify. RFC 1847 section 2.1 asks for the first body part
as received.

The body was never the problem: `LazyParser::encode()` hands back the received
buffer while the part is not dirty. Only the header block was rebuilt.

`MultipartMixedContents` now keeps a read-only view of the first part exactly as
it arrived, header block included and the CRLF that belongs to the following
boundary excluded, and `checkSignature()` verifies against that. What is kept is
a `Data::Share` into the buffer being parsed, so parsing an ordinary multipart
body costs nothing extra; the accessor hands back a copy of it. Only the first
part is kept, the one the signature covers. Where there is no view, for a body
built locally rather than parsed, verification falls back to the previous
behaviour.

One more change is needed for that to take effect in DUM.
`EncryptionManager::Decrypt::decrypt()` calls `isEncrypted()` before
`getContents()`, and `isEncryptedRecurse()` reached for the mutable `parts()`
just to see what kind of body it had. The non-const `checkParsed()` marks the
object DIRTY, so the received bytes were already gone when `checkSignature()`
ran. It now inspects through the const overload and writes back only when the
recursion replaced a part.

## Measured

Signing the exact bytes of the first part with OpenSSL, the same flags
`BaseSecurity::sign` uses, then letting resiprocate parse and verify:

| case | before | after |
| --- | --- | --- |
| header block exactly as resiprocate writes it (control) | verifies | verifies |
| plus `MIME-Version: 1.0` | `SignatureIsBad` | verifies |
| `Content-Disposition` before `Content-Type` | `SignatureIsBad` | verifies |
| two spaces after the colon | `SignatureIsBad` | verifies |
| `Content-Type: text/plain; charset=us-ascii` | `SignatureIsBad` | verifies |
| header block of the capture in `SipTortureTests.cxx` | `SignatureIsBad` | verifies |
| header order of that capture alone | `SignatureIsBad` | verifies |

The control row matters: it says the harness can tell success from failure.
6 of 6 differing layouts go from `SignatureIsBad` to verifying while the control
keeps verifying on both sides. Taking the change back out reproduces the 6
failures exactly, so the harness follows the production path rather than its own
copy of it.

An eighth case is not in the table: a part carrying a header that
`Headers::getType` does not know does not reach verification at all, in either
state. It is described under Limitations.

The DUM half is measured indirectly. The harness does not drive
`EncryptionManager`; it runs the same body twice from the same build, once with
the const inspection and once with the mutable `parts()` call
`isEncryptedRecurse()` used to make, so what is measured is the access mode
rather than that function:

| inspection before `checkSignature()` | result over the 8 cases |
| --- | --- |
| const overload, as it is after this change | 7 verify, 0 fail, 1 does not reach verification |
| mutable `parts()`, as `isEncryptedRecurse()` did before | 1 verifies, 6 fail, 1 does not reach verification |

The single one that still verifies in the lower row is the control, whose header
block happens to match what resiprocate itself writes.

Regression side, a second harness against the same build:

- `sign()` then `checkSignature()` on the same object, which has no view and
  takes the fallback: still verifies.
- `sign()`, encode to the wire, parse again, verify, which is what two
  resiprocate instances actually do: still verifies, and the view is used.
- Body altered after signing: still rejected, so verification did not simply
  become permissive.

Suite: 69 of 69 with assertions on, and 69 of 69 again inside
`debian:bullseye-slim`, the image `.github/workflows/linux-ci.yml` names, built
there with gcc-10 against OpenSSL 1.1.1w. cmake comes from pip in that run
rather than from the workflow's own upgrade step, and the debify package set is
not installed, so the run reproduces the compiler and the headers rather than
the CI job. Both local runs
come from a build with the optional components off (recon, repro, reflow,
return, media, rend, tfm), so they cover the stack and its tests and not the
whole tree.

clang-tidy with the checks in `.clang-tidy` reports nothing on the four changed
source files, run inside that same image with clang-13, which is the leg where
the workflow calls `run-clang-tidy-13`.

I ran the earlier measurements in RelWithDebInfo at first, which sets `-DNDEBUG`
and silences `resip_assert`. Everything above is from a build with assertions
on.

## New test

`resip/stack/test/testMultipartRawParts` checks the contract: the first part
comes back unchanged for the header layouts above, a differing part reads back
differently so the check cannot pass on a stub, const access keeps the view
while mutable access drops it, and a copy or a locally built body has none. It
needs no certificates and no arguments.

The crypto level measurements above come from two harnesses that need a key
pair, so they are not in the suite.

## Limitations

- `MultipartMixedContents` grows by one `Data` member, and `debian/` packages a
  shared library while `SO_ABI` is commented out in `CMakeLists.txt`, so nothing
  raises the soname. If you would rather not grow the class, the same bytes can
  be derived on demand from the buffer the object was parsed from; that needs a
  const accessor for it in `LazyParser`, which is why I did not start there.
  Happy to switch if you prefer that shape.
- A caller that takes mutable access through `parts()` before verifying gets the
  old behaviour rather than a wrong answer. All 4 non-test call sites reach
  `checkSignature` without doing that: `TuIM.cxx:289` and `TuIM.cxx:607` call it
  directly, `EncryptionManager.cxx:1219` after the change above, and the
  `InfoLog` preceding `Helper.cxx:2075` is const and does not mark the body
  dirty. The calls under `test/` sign a locally built body, which carries no
  view either way.
- The branch in `isEncryptedRecurse()` that writes a replaced part back is not
  reached by any test in the tree, so the change above rests on reading rather
  than on a run. The table further up measures the access mode, not that
  function.
- Parts carrying headers that `Headers::getType` does not know are out of scope
  for this change and are not covered by the new test. That is older behaviour
  and I did not touch it.
- `checkSignature()` does not pass `PKCS7_BINARY` while `sign()` does. I left
  that alone since it is not needed for the cases here, but it looks like an
  asymmetry worth a separate look.

## Commits

```
1) resip/stack: record the first body part as it arrived
2) resip/stack: verify multipart/signed against the received bytes
3) resip/dum: inspect a signed body without dropping the received bytes
4) resip/stack/test: add testMultipartRawParts
```


sgodin:
Hi @Nexory - thanks for another fix!

I've asked Claude to review and here is the response:

Nice work — this is the right fix. Verifying against the received octets is what RFC 1847 requires, and the re-encode was never going to interoperate with anything that didn't order and space its headers exactly the way we do.

I traced the byte alignment and it checks out: skipToChars(boundary) lands on the CRLF that belongs to the next delimiter, so the captured span is exactly what encodeParsed() puts on the wire and exactly what sign() signs. The Data::Share handling is right too — the move-assign through takeBuf() keeps it zero-copy, and getRawFirstPart() returning by value goes through initFromString(), so the copy really is self-contained. And the parts().empty() guard in isEncryptedRecurse fixes a latent *begin()-on-empty that's been sitting there for years. Good catch.

Four things before I merge.

1. Helper::getSdpRecurse() has the same problem you fixed in EncryptionManager. resip/stack/Helper.cxx:2250 reads the parts through the non-const overload:

MultipartSignedContents::Parts::const_iterator it = mps->parts().begin();

mps is non-const, so that resolves to the mutable parts() and wipes mRawFirstPart. Any stack-level UA that peeks at the SDP before calling checkSignature() silently gets the old behaviour back. DUM users are fine — EncryptionManager runs ahead of InviteSession in the feature chain — but direct Helper::getSdp() callers aren't. Same treatment as isEncryptedRecurse: bind a const MultipartSignedContents* and read through that. While you're in there, that begin() is dereferenced without an empty() check, so it wants the same guard you added in the other spot.

2. Please log the fallback in checkSignature(). Right now an empty rawFirst silently reverts to the path this PR exists to replace, and "locally built body" and "someone upstream called parts()" are indistinguishable. A DebugLog/InfoLog in the else branch is what makes the whole design maintainable — it turns a future regression from "signatures mysteriously stopped verifying" into something greppable. Given that parts() now has a security-relevant side effect that isn't visible at the call site, I'd like that safety net in place.

3. The copy-semantics paragraph on getRawFirstPart() isn't accurate. It says a copy never has a record because it "is not parsed again" — but LazyParser's copy ctor preserves mState and HeaderFieldValue's copy ctor deep-copies and takes ownership. So a copy of a body that was still NOT_PARSED will parse on first access and establish a perfectly valid view into its own buffer. The behaviour is fine either way; it's the doc that's backwards. The test only covers the already-parsed case, so it doesn't catch it — worth adding a case that copies before the source is touched.

4. Trivial: the test is linked against TestSupport.cxx and includes rutil/DataStream.hxx, and doesn't use either.

On clear() — you're right that it deletes the parts and leaves the vector populated, and the parse() comment describes it exactly. That's a real double-free through operator=, but it predates this PR and I'd rather not widen the scope here. Either add the one-line mContents.clear() if you want it done, or leave it and I'll open a separate issue.

ABI isn't a concern; this is going into master.

Nexory:
Thanks for the review, that is a lot of useful detail. All four are addressed, in three
commits that map to your numbering, and the branch is rebased onto current master since it
had drifted six commits behind and two of those touched files this PR changes.

1. `Helper::getSdpRecurse()` now binds a `const MultipartSignedContents*` and reads through
that, with the same comment as in `isEncryptedRecurse`, and the `empty()` guard before the
first part.

I measured that site rather than take the shape for granted, since a call path that looks
identical is not always reachable. With the old `Helper.cxx` and everything else in place,
a signed body whose first part is an SDP has its record before `Helper::getSdp()` and an
empty one after. With the change the record survives the call unchanged. `getSdp()` returns
the SDP either way, so the const read costs nothing at that call site.

2. `checkSignature()` logs which path it took when nothing was recorded, and names both
cases so the line tells you which one you are looking at.

3. You are right and the test proves it: a copy taken while the source is still unparsed
does end up with a record, because the copy constructor deliberately does not force the
original to parse and the copy then parses its own buffer. The note now describes both
cases, and there is a case that copies before anything reads the original, checking that
the copy reports the same bytes and that the original still works afterwards.

4. `TestSupport.cxx` and `rutil/DataStream.hxx` are gone from the test.

On `clear()`: leaving it out as you suggested. If you open the issue I will send the patch
for it once this one is merged, so it does not compete for review attention here.

Measured on the rebased branch: the interop harness goes from one variant verifying to
seven of eight, the eighth being the pre-existing abort in `Contents.cxx` that fails the
same way in both states. Full suite 71 of 71 locally and 71 of 71 in `debian:bullseye-slim`
built with gcc-10 against OpenSSL 1.1.1w, which is the combination that broke a previous
change of mine in this area. clang-tidy reports no project findings. Removing the fix puts
the harness back to one of eight, so the tests are not passing on their own account.

sgodin:
Thanks, this addresses everything. Re-reviewed the three new commits.

The Helper::getSdpRecurse fix is right — const read, empty() guard before front(), and the recursion only dirties the child so the parent's record survives. The fallback log does its job. And the new copy test is better than the one it replaces: asserting that both the copy and the original independently produce the correct bytes is what actually pins the HeaderFieldValue deep-copy ownership the whole lifetime story depends on.

One thing to fix before I merge: the copy-constructor comment in MultipartMixedContents.cxx:56 is now stale. It still says the copy "is never parsed again … A copy therefore reports no raw bytes and checkSignature() falls back" — exactly the claim you corrected in the header. operator= points a reader straight at it ("see the copy constructor for why the view is not carried over"), so as it stands there are two comments asserting opposite things and you have to go read LazyParser to break the tie. That's the cost the header rewrite was meant to remove.

Two optional nits while you're in there:

- The new log says the fallback is "expected for a locally built body and a regression if the body was parsed." A copy of an already-parsed body is parsed, has no record, and is perfectly normal — SipMessage's copy ctor clones mContents, so it's reachable. Three cases, not two, if you want the log to stay honest.
- "A copy is not empty by construction" reads as "a copy is not empty," which is backwards from what you mean. Something like "whether a copy has a record depends on when it was taken" says it directly.

Nexory:
Fixed. It turned out to be three places rather than one, so let me say what changed and why.

The copy constructor comment now says what the header says. The view is not carried over, since
it points into the buffer of the original; a body copied before anything reads it stays unparsed,
holds a deep copy of the buffer, and records a view into that buffer on first access; a body
copied after it was parsed keeps the parts and no record, and `checkSignature()` falls back for
that one.

`operator=` carried the same stale claim one function further down. Its comment said the record
"stays empty afterwards" and sent the reader to the copy constructor for the reason, so fixing
only the copy constructor would have moved your objection instead of answering it. I measured
that rather than assume it: assigning from a source nothing has read leaves this object unparsed
with a buffer of its own, and its first access records a view into that buffer, byte for byte the
same as the copy constructor gives. Only assignment from an already parsed source leaves no
record. The comment now says that, and two cases pin both directions. Gutting `getRawFirstPart()`
turns the first of them red, so it covers the line rather than passing on its own account.

Both nits taken.

The log names three cases now and calls only the last one a defect. You are right that a copy of
an already parsed body is ordinary rather than a regression, and it is reachable without anyone
doing anything wrong: `SipMessage::init()` runs `mContents = rhs.mContents->clone()` at
`resip/stack/SipMessage.cxx:263` whenever the body has been parsed, so a plain message copy lands
there. `testMultipartRawParts` already pinned that case, so the log and the test now agree.

The header sentence opens with "Whether a copy has a record depends on when it was taken", which
is what the rest of the paragraph goes on to explain.

No production behaviour is touched: four comments, the wording of one log line, and two test cases.

Suite is 71 of 71 locally and 71 of 71 in `debian:bullseye-slim` at
`sha256:f313b4bd62667092a59b3a664d7d3ab8b5e65f41675f48e81455a15dc5abe792` with gcc-10 against
OpenSSL 1.1.1w, and clang-tidy reports no project findings.

Nexory:
The clang-13 job is red on this branch: `72 - sanityTests (SEGFAULT)`. gcc-10 and MSVC are green
on the same commit and clang-13 is green on master, so the obvious reading is that this branch
causes it. I could not confirm that.

I built both states in your pinned `debian:bullseye-slim`, clang 13.0.1 from the same debify
repository, cmake 3.27.9, TFM and repro on. Branch against master, three ways:

- `sanityTests` alone: both pass, 2299.42 s against 2298.86 s.
- Under AddressSanitizer: no memory error on either side, the same 7 failures in the same
  positions. 510 `__asan` symbols in the binary, so it really was linked in.
- Full suite: byte-identical progress bands, 288 characters, 17 failures, the same names, exit 8
  on both.

6 runs, no difference anywhere. That is the interesting part for the lifetime question your review
circled: `mRawFirstPart` is a `Data::Share` view into the parse buffer, exactly what ASAN would
catch if it outlived that buffer, and ASAN is silent.

Not a clean bill of health, though, because my build is not yours: recon, sipxtapi, the
database and SNMP bindings and the clang-tidy pass are off, and it fails 17 tests on
master that your CI passes.

Also a correction. My last comment said 71 of 71. That is 71 of the 77 your CI runs, because the
profile I measure with builds without TFM and repro, so `sanityTests` was never in it.

If you can re-run the clang-13 job on this commit, that is the one measurement I cannot make.

sgodin:
@Nexory I suspect the CI error was something intermittent - I've re-run it.  I will merge after a clean run.  Thank you for this work.

PR #485, opened 2026-08-30
Add ordered view over `SipMessage::KnownHeaders`

Lastique:
### Description

This PR implements an ordered view over known headers in a `SipMessage`. The view provides read-only access to headers in the order of `Headers::Type` enum values, which is useful for deterministic serialization of the message.

The view relies on the bit mask of non-"unused" entries in `KnownHeaders`. The bit mask is transparently maintained when elements are inserted and removed from the container. On iteration, the bit mask is scanned for set bits indicating valid headers.

This bit mask duplicates information that was already present in the `HeaderInfo` struct, as the special value of `mType == Headers::UNKNOWN`. The special value of `mType` and the associated logic is preserved to keep iteration over the unordered headers fast, as it keeps memory accesses localized in the `mHeaders` vector.

Use this ordered view in `SipMessage::encode` and `encodeEmbedded` instead of looping over all known header types. This significantly optimizes serialization performance.

### Background

I have implemented two versions of code that produces ordered headers in `encode` and `encodeEmbedded` that were discussed in comments to commit 7a9b42395df800b032904fbe2581ca91ed54ebb9. This PR represents @sgodin's idea of using a bit mask to optimize iteration over SIP headers in the ordered way, plus the ordered view as a way to expose this feature. The other idea, which is to order SIP headers in the container on insert, is implemented in https://github.com/Lastique/resiprocate/tree/feature/ordered_known_headers (specifically, the last two commits: [1](https://github.com/Lastique/resiprocate/commit/c73ee9706cf059338b25737ad378992878220145), [2](https://github.com/Lastique/resiprocate/commit/08f54a8a4e382601f0ef631647b4423ee7dd3070)).

I have benchmarked the two designs and have decided to submit the version with the bit mask as a PR after all. If the other version is still preferred, I can submit that as a separate PR instead.

Note that the "ordered insert" version has two commits. The first one makes all SIP header inserts via `SipMessage` API ordered. The second one adds unordered inserts and makes `MsgHeaderScanner` use that API to recover some lost deserialization performance from the first commit, which made all header inserts more expensive. This made `SipMessage` API contract more complex, as the user would have to sort headers if he used unordered APIs, before he uses ordered APIs. See the commit messages for more details.

### Benchmark

Benchmark code:

[resip_sipmessage_benchmark.cpp](https://github.com/user-attachments/files/31622397/resip_sipmessage_benchmark.cpp)

Compile with:

```
g++ -O3 -std=gnu++17 -o resip_sipmessage_benchmark resip_sipmessage_benchmark.cpp -lresip -lrutil
```

Add libresiprocate include and library paths, as needed.

The benchmark tests three operations:

- Deserialize SIP message
- Add a header to that message (which should roughly be in the middle of headers, if ordered by `Headers::Type`)
- Serialize that message

These operations are tested on two versions of an input SIP INVITE message: good and bad. The good one has headers already ordered by `Headers::Type`, the bad one has the reverse order. Otherwise, the messages are the same. For each operation, an average time over 100k operations was measured.

Five versions of libresiprocate were benchmarked:

- 1.14.0, which uses unordered SIP headers internally and on serialization.
- 7a9b42395df800b032904fbe2581ca91ed54ebb9, which changed serialization to loop over `Headers::MAX_HEADERS` instead of the headers present in the message.
- https://github.com/Lastique/resiprocate/commit/c73ee9706cf059338b25737ad378992878220145, which orders headers on insert, but does not yet optimize `MsgHeaderScanner`.
- https://github.com/Lastique/resiprocate/commit/08f54a8a4e382601f0ef631647b4423ee7dd3070, which is the full "ordered insert" version.
- This PR, which is the bit mask version.

Each libresiprocate version was compiled with:

```
cmake -DCMAKE_BUILD_TYPE=Release -DBUILD_TFM=0 -DBUILD_RECON=0 -DBUILD_REPRO=0 -DBUILD_RETURN=0 -DBUILD_REFLOW=0 -DBUILD_REND=0 -DBUILD_MEDIA=0 -DBUILD_PYTHON=0 -DBUILD_QPID_PROTON=0 -DUSE_MYSQL=0 -DASIO_INCLUDE_DIR=<asio include> -DBoost_INCLUDE_DIR=<boost include> -DSRTP2_INCLUDE_DIR=<libsrtp include> -DSRTP2_LIBRARY=srtp2 ..
make -j8
```

Test platform: Kubuntu 24.04, gcc 13.3.0, Intel Core i7-4700HQ.

Test results for good message (all numbers are nanoseconds per operation, lower is better):

|             | 1.14.0 | `MAX_HEADERS` loop | Ordered insert, no scanner | Ordered insert + scanner | Bitmask |
|-------------|--------|--------------------|----------------------------|--------------------------|---------|
| Deserialize |   3210 |               3116 |                       3152 |                     3323 |    3242 |
| Add header  |     55 |                 49 |                         64 |                       98 |      54 |
| Serialize   |   2049 |               2375 |                       2079 |                     2183 |    2160 |

Bad message:

|             | 1.14.0 | `MAX_HEADERS` loop | Ordered insert, no scanner | Ordered insert + scanner | Bitmask |
|-------------|--------|--------------------|----------------------------|--------------------------|---------|
| Deserialize |   3199 |               3089 |                       3119 |                     3275 |    3161 |
| Add header  |     55 |                 50 |                         63 |                       96 |      53 |
| Serialize   |   2048 |               2351 |                       2058 |                     2147 |    2093 |

A few notes on test results:

- The single header addition operation is rather fast on its own, so large swings from one run to another are possible. However, in general, I noticed that the "ordered insert" versions tend to produce larger times in this test (as expected).
- The full "ordered insert" version, with `MsgHeaderScanner` optimization, produced somewhat larger times than the one without the `MsgHeaderScanner` optimization, which is unexpected. The optimization only affected the deserialization, and I cannot explain the effect on the other two operations. I've also tested debug builds of libresiprocate, and there the results were as expected - the full "ordered insert" version is slightly faster in deserialize, more or less the same in the other two operations. Since the bitmask version is faster anyway, I did not investigate further.
- The effect of a bad input message is lower than I anticipated, almost nonexistent. It may be because the good message is tested first and the bad one after than, so it is possible that some caches are warm when the bad message is tested.
- The `MAX_HEADERS` loop is consistently and significantly slower in serialize than the other versions. The "ordered insert" versions have slower deserialize and header inserts than 1.14.0, but serialize is the same. The bitmask version is the same as 1.14.0 in deserialize and header inserts and slightly slower in serialize (definitely not as much as the `MAX_HEADERS` loop).

### DIFF
--- a/resip/stack/SipMessage.cxx
+++ b/resip/stack/SipMessage.cxx
@@ -29,9 +29,23 @@ using namespace std;
 void
 SipMessage::KnownHeaders::clear() noexcept
 {
-   for (iterator it = begin(), e = end(); it != e;)
+   if (mSize > 0)
    {
-      erase(it++);
+      for (reference elem : mHeaders)
+      {
+         if (elem.getType() != Headers::UNKNOWN)
+         {
+            elem.getValues()->clear();
+            elem.setType(Headers::UNKNOWN);
+         }
+      }
+
+      for (UsedBitMaskWord& word : mUsedBitMask)
+      {
+         word = 0u;
+      }
+
+      mSize = 0u;
    }
 }
 
@@ -46,7 +60,7 @@ SipMessage::KnownHeaders::clearAndDispose(Disposer&& disposer) noexcept
    }
 
    mHeaders.clear();
-   mSize = 0;
+   mSize = 0u;
    resetIndices();
 }
 
@@ -54,12 +68,17 @@ SipMessage::KnownHeaders::clearAndDispose(Disposer&& disposer) noexcept
 void
 SipMessage::KnownHeaders::erase(iterator it) noexcept
 {
-   resip_assert(it->getType() != Headers::UNKNOWN);
+   const Headers::Type type = it->getType();
+   resip_assert(type != Headers::UNKNOWN);
    resip_assert(it->getValues() != nullptr);
    resip_assert(mSize > 0);
 
    it->getValues()->clear();
    it->setType(Headers::UNKNOWN);
+
+   mUsedBitMask[static_cast<unsigned int>(type) / UsedBitMaskWordBits] &=
+      ~(static_cast<UsedBitMaskWord>(1u) << (static_cast<unsigned int>(type) % UsedBitMaskWordBits));
+
    --mSize;
 }
 
@@ -93,15 +112,25 @@ SipMessage::KnownHeaders::insert(Headers::Type type, ValuesFactory&& valuesFacto
       ++mSize;
    }
 
+   mUsedBitMask[static_cast<unsigned int>(type) / UsedBitMaskWordBits] |=
+      static_cast<UsedBitMaskWord>(1u) << (static_cast<unsigned int>(type) % UsedBitMaskWordBits);
+
    return iterator(mHeaders.begin() + pos, mHeaders.end());
 }
 
-/// Resets all header indices to `InvalidHeaderIndex`
+/// Resets all header indices to `InvalidHeaderIndex` and clears the "used" bit mask
 void
 SipMessage::KnownHeaders::resetIndices() noexcept
 {
    for (HeaderIndex& index : mHeaderIndices)
+   {
       index = InvalidHeaderIndex;
+   }
+
+   for (UsedBitMaskWord& word : mUsedBitMask)
+   {
+      word = 0u;
+   }
 }
 
 /// Finds header information, if present in the list. Returns `end()` if not found. Does not produce "unused" entries.
@@ -820,29 +849,11 @@ SipMessage::encode(EncodeStream& str, bool isSipFrag) const
 #endif
    }
 
-   // Iterate by header type enum value (not insertion order) so that headers are
-   // encoded in the deterministic, enum-declared order. See the comment in
-   // HeaderTypes.hxx: "The Type enum controls the order of output of the headers
-   // in the encoded SipMessage". The leading entries of that enum follow the
-   // recommendation in RFC 3261 section 7.3.1, that headers needed for proxy
-   // processing appear towards the top of the message to facilitate rapid parsing.
-   //
-   // !slg! Note this costs more than iterating mKnownHeaders directly: find() is
-   // O(1), but the loop still probes all Headers::MAX_HEADERS types no matter how
-   // few headers the message actually carries, whereas iterating the container
-   // visits only the headers that are present. Recovering the enum order without
-   // paying that is worth revisiting.
-   for (int t = 0; t < Headers::MAX_HEADERS; ++t)
+   for (KnownHeaders::OrderedView::const_reference info : mKnownHeaders.ordered())
    {
-      const Headers::Type type = static_cast<Headers::Type>(t);
-      if (type == Headers::ContentLength) // !dlb! hack...
+      if (info.getType() != Headers::ContentLength) // !dlb! hack...
       {
-         continue;
-      }
-      auto it = mKnownHeaders.find(type);
-      if (it != mKnownHeaders.end())
-      {
-         it->getValues()->encode(type, str);
+         info.getValues()->encode(info.getType(), str);
       }
    }
 
@@ -878,31 +889,21 @@ EncodeStream&
 SipMessage::encodeEmbedded(EncodeStream& str) const
 {
    bool first = true;
-   // Iterate by header type enum value (not insertion order) so that headers are
-   // encoded in the deterministic, enum-declared order. See encode() above, including
-   // the note about the cost of probing all Headers::MAX_HEADERS types.
-   for (int t = 0; t < Headers::MAX_HEADERS; ++t)
+   for (KnownHeaders::OrderedView::const_reference info : mKnownHeaders.ordered())
    {
-      const Headers::Type type = static_cast<Headers::Type>(t);
-      if (type == Headers::ContentLength)
+      if (info.getType() != Headers::ContentLength)
       {
-         continue;
+         if (first)
+         {
+            str << Symbols::QUESTION;
+            first = false;
+         }
+         else
+         {
+            str << Symbols::AMPERSAND;
+         }
+         info.getValues()->encodeEmbedded(Headers::getHeaderName(info.getType()), str);
       }
-      auto it = mKnownHeaders.find(type);
-      if (it == mKnownHeaders.end())
-      {
-         continue;
-      }
-      if (first)
-      {
-         str << Symbols::QUESTION;
-         first = false;
-      }
-      else
-      {
-         str << Symbols::AMPERSAND;
-      }
-      it->getValues()->encodeEmbedded(Headers::getHeaderName(type), str);
    }
 
    for (UnknownHeaders::const_iterator i = mUnknownHeaders.begin(); 
--- a/resip/stack/SipMessage.hxx
+++ b/resip/stack/SipMessage.hxx
@@ -3,10 +3,12 @@
 
 #include <sys/types.h>
 
+#include <cstddef>
 #include <list>
 #include <vector>
-#include <utility>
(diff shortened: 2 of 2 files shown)


resip/stack/SipMessage.cxx
```diff
@@ -29,9 +29,21 @@ using namespace std;
...
+      {
+         if (elem.getType() != Headers::UNKNOWN)
+         {
+            elem.getValues()->clear();
+            elem.setType(Headers::UNKNOWN);
+         }
+      }
+
+      for (UsedBitMaskWord& word : mUsedBitMask)
+         word = 0u;
```
sgodin, line 45: Minor - please curly brace this.  There are a few others in this file as well further down (ie: resetIndices).

sgodin:
@Lastique  Thanks for benchmarking both — the data is much more convincing than the
reasoning either of us was doing, and I'm glad you measured rather than took
my guess. Happy with the bitmask version. A couple of things before it lands.

**Blocker: this won't compile on 32-bit MSVC.**

`countrZeroNz(unsigned __int64)` guards only the `_tzcnt_u64` path with
`_M_AMD64` and falls back to `_BitScanForward64`. But MSVC declares
`_BitScanForward64` only for x64 and ARM64 — in `intrin0.inl.h` it's wrapped in
`__MACHINEX64` / `__MACHINEARM64`, and `__MACHINEX64` is redefined to the
discard macro whenever `_M_X64` is undefined. So on `_M_IX86` that `#else`
branch calls an undeclared function. Since `Iterator` is a plain nested class,
the overload is compiled whether or not it's ever called, so including
SipMessage.hxx is enough to break the build. We still ship Win32 configurations
in the vcxproj files, so this would bite.

Simplest fix is to just not declare the overload where it can't work — with
`UsedBitMaskWord` being `std::size_t`, it's unreachable on 32-bit anyway:

    #if defined(_M_X64) || defined(_M_ARM64)
                            static unsigned int countrZeroNz(unsigned __int64 bitmask) noexcept
                            {
    #if defined(_M_AMD64)
                               return static_cast<unsigned int>(_tzcnt_u64(bitmask));
    #else
                               unsigned long res;
                               _BitScanForward64(&res, bitmask);
                               return static_cast<unsigned int>(res);
    #endif
                            }
    #endif

Alternatively, mirror what the MSVC STL does in `__msvc_bit_utils.hpp` and split
the 64-bit value into two 32-bit `_BitScanForward` calls on x86. That keeps the
overload usable if `UsedBitMaskWord` ever stops tracking `size_t`, at the cost of
a few more lines. Either is fine by me.

**Minor:** worth confirming `#pragma intrinsic(_tzcnt_u32)` doesn't produce
C4163 — I'm not sure `_tzcnt_u32` is on the list of names that pragma accepts,
and it'd be noisy if we build that config with /WX. `_BitScanForward` is
definitely fine.

Otherwise this looks good to me. Sizing the mask from `MAX_HEADERS` rather than
hardcoding a word count addresses the thing I was worried about — we're at 97
today and creeping up, and this degrades to an extra word rather than a cliff.
Every mutation path looks covered (insert, erase, clear, clearAndDispose via
resetIndices). And thanks for fixing `UsedIterator::operator++` to return a
reference while you were in there.

One note on your results for the record: the `MAX_HEADERS` loop I introduced
cost ~16% on serialize (2375 vs 2049 ns), which is quite a bit worse than I
guessed when we started this. Your bitmask version gets about two thirds of
that back on the good message and ~85% on the bad one, at no measurable cost to
deserialize or header insert. That's a better outcome than I expected from
either design.

sgodin:
**Suggestion: move `countrZeroNz` (and the `<intrin.h>` preamble) to rutil/compat.hxx.**

Two reasons beyond tidiness. First, the `#include <intrin.h>` and the
`#pragma intrinsic` block currently sit at the top of SipMessage.hxx, so every
MSVC TU that touches a SipMessage pulls in the Windows intrinsics header —
compat.hxx is where we already hide that kind of thing. Second, a helper buried
in a private nested class can't be unit tested; in rutil it can get a test in rutil/test
and the platform guards live in one place.

compat.hxx is already in SipMessage.hxx's include graph (via rutil/Data.hxx), so
this costs nothing, and it's already listed in rutil/CMakeLists.txt so there's no
vcxproj churn. It's header-only — no compat.cxx exists and none is needed.

There's also a precedent there for retiring the shim automatically. compat.hxx
already does:

    #if __cplusplus >= 202002L || (defined(_MSVC_LANG) && _MSVC_LANG >= 202002L)
    #  define RESIP_LAMBDA_CAPTURE_ALL_AND_THIS =, this
    #else
    #  define RESIP_LAMBDA_CAPTURE_ALL_AND_THIS =
    #endif

so the helper could lead with a `std::countr_zero` path guarded on
`__cpp_lib_bitops`, and the intrinsic branches become dead weight the day we
raise the baseline. Keeping the name close to `countr_zero` makes that
migration obvious.

The C4146 pragma should stay where it is — that's about the unary minus in
advanceToNextUsedEntry, not about the intrinsic.

Entirely your call whether to fold this into this PR or do it as a follow-up.

Lastique:
> **Minor:** worth confirming `#pragma intrinsic(_tzcnt_u32)` doesn't produce
> C4163 — I'm not sure `_tzcnt_u32` is on the list of names that pragma accepts,
> and it'd be noisy if we build that config with /WX.

It's [fine](https://msvc.godbolt.org/z/d34xv7csb) at least with VS 16.0.

> Suggestion: move `countrZeroNz` (and the `<intrin.h>` preamble) to rutil/compat.hxx.

I can move `countrZeroNz` there no problem, but consider that `compat.hxx` is included in many other headers, none of which currently need `intrin.h` or `countrZeroNz`. Also note that in order to make `countrZeroNz` generic I will need to include `<limits>` and probably `<type_traits>` as well, which are currently not included in `compat.hxx`. I understand that you want to localize portability stuff in one place, but such "kitchen sink" kind of headers that get included everywhere tend to carry an increasing cost of inclusion. This cost may be justified if you plan to use `countrZeroNz` in multiple (many) places, although even then I would probably try to avoid it.

Just letting you know of the consequences. Do you still prefer `countrZeroNz` moved there?

> the helper could lead with a `std::countr_zero` path guarded on
> `__cpp_lib_bitops`, and the intrinsic branches become dead weight the day we
> raise the baseline.

`countrZeroNz` is different from `std::countr_zero` in that it requires the input value to be non-zero. This allows it to be compiled to a single `bsf` instruction on x86 CPUs without BMI1, whereas `std::countr_zero` would require an extra branch on such targets. That branch would be unnecessary if the input is already guaranteed to be non-zero by the caller. Which is the case in `SipMessage`. Yes, this is just a micro-optimization, but still there's no reason to use `std::countr_zero` here, even if available.

If you ever switch to C++20 and `std::countr_zero`, and consider this micro-optimization not worth keeping, you can simply remove `countrZeroNz` and replace its uses with `std::countr_zero`.

I will update the code according to your other comments.

sgodin:
@Lastique  Excellent points on my compat.hxx suggestion.  I agree we don't need that change.

Thanks for checking on the pragma - good, nothing to do there.

I also agree your points on std::countr_zero - thanks for the correction.

So just the 32-bit MSVC fix (and minor bracketing changes) from my earlier comment. Thanks for bearing with the review.

Lastique:
I have updated the code, but I have no way of testing MSVC for 32-bit targets.

sgodin:
I verified the fix against MSVC 14.51 (VS 18):

- x86 target, /W4: compiles clean (previously `error C3861`)
- x64 target, /W4: compiles clean
- split-scan fallback: correct for all 64 single-bit positions and for each with
  an additional high bit set

LGTM.  I will merge when the CI stuff is done it's thing.  :)    Thanks!

PR #487, opened 2026-09-11
Let a transport carry a DSCP class on its sockets

mykhailopopivniak:
An application that wants its SIP signalling in a particular DSCP class
cannot reach the sockets the stack owns. AfterSocketCreationFuncPtr is
the only hook, so every application repeats the same setsockopt and
chooses between IP_TOS and IPV6_TCLASS itself. Re-running it over the
sockets already open, through invokeAfterSocketCreationFunc(), walks the
connection list from the transaction controller's thread, while a
RESIP_TRANSPORT_FLAG_OWNTHREAD transport services that list from its
own.

Give the value to Transport instead. Every site that configures a new
socket routes through one place, so the listener and every connection on
it are covered, UDP included.

The class rather than the whole ToS byte, because the low two bits are
the ECN bits and no two platforms treat an application that writes them
alike. Measured: a datagram socket on Linux keeps them, a stream socket
clears them, and Windows rejects the value with WSAEINVAL. A class of 0
to 63 behaves the same on all three.

A changed value arrives on whichever thread reads the configuration, so
setDscp() only raises a request and each transport serves it from its
own process() cycle. onReload() and the existing
invokeAfterSocketCreationFunc() path are untouched.

The listener read-back reports the state of the marking rather than a
transition, so a failure is re-stated on every reload instead of
scrolling away.

The range is checked in the setter because checking it per socket would
turn one mistyped value into an error line per accepted connection.

A transport with no class to write and no earlier mark to undo is left
alone. Clearing the value to -1 writes 0 once, so a connection that
outlives the change does not keep the old marking.


sgodin:
@mykhailopopivniak Thanks for this. Having one place that marks every socket a transport configures, with the refresh served from the transport's own thread, is a good fit for the stack.  I had Claude analyze the PR and it had these suggestions.

I built and ran this on both Linux (Ubuntu 22.04, GCC) and Windows 11 (MSVC, CMake + Ninja). All three tests pass on Linux, and the refresh also works against a real `SipStack` with UDP and TCP transports: setting a byte marks the listeners and existing connections, and clearing it writes 0 back. Windows is where the problems are.

## 1. Windows CMake build is broken in rutil

`rutil/Socket.cxx` lines 312 and 331 call `strerror(e)`. MSVC flags that as C4996, and our top-level `CMakeLists.txt` adds `/sdl` for MSVC, which makes it an error, so `rutil` itself doesn't compile:

```
rutil\Socket.cxx(312): error C4996: 'strerror': This function or variable may be unsafe...
```

`strerror` also doesn't know Winsock error codes, so the message wouldn't help on Windows anyway (even `strError()` prints "Unknown error" for WSAEINVAL). Logging the number is enough:

```cpp
int e = getErrno();
ErrLog(<< "setsockopt(" << optionName << ", " << tos << ") failed on fd " << fd << ": error " << e);
```

## 2. The tests don't build or pass on Windows

The tests are registered unconditionally, so with `BUILD_TESTING` on (the default) the Windows build fails here too:

- **Winsock takes `char*` for option values.** `testSocketTos.cxx` lines 43, 49, 54 and `testTransportTos.cxx` lines 152 and 174 pass `int*`, which fails with C2664. Casting to `(const char*)` / `(char*)` works on both platforms, the same way `Socket.cxx` already does it.
- **No `initNetwork()`.** `testSocketTos` and `testTransportTos` open sockets before anything calls WSAStartup, so every `::socket()` returns `INVALID_SOCKET` on Windows. (`testTransportTosRefresh` works because `SelectInterruptor` calls it.) Calling `initNetwork()` at the top of `main()` fixes this.
- **Some checks are Linux-only.** With the two fixes above, 4 checks in `testSocketTos` still fail on Windows:
  - `V6: left the IPv4 level untouched`: on Windows, `setsockopt(IP_TOS)` on an `AF_INET6` socket fails with WSAEINVAL, and `getsockopt(IP_TOS)` returns the traffic class, so the decoy never lands.
  - `datagram: keeps the ECN bits it was given`, `stream: reports its own ECN bits`, and, as a consequence, `a refused value leaves the socket as it was`. See the next section.

With those fixes, `testTransportTos` and `testTransportTosRefresh` pass on Windows.

## 3. The ECN bits behave differently on every platform, so I'd suggest taking DSCP instead of the raw byte

I probed `IP_TOS` / `IPV6_TCLASS` with and without the low two bits set:

| value | Linux UDP | Linux TCP | Windows UDP + TCP (v4 and v6) |
|---|---|---|---|
| 160, 184, 252 (ECN bits clear) | kept | kept | kept |
| 161, 162, 163, 255 (ECN bits set) | kept | ECN bits silently dropped | **rejected, WSAEINVAL** |

So a configured byte of 161 marks UDP as 161 and TCP as 160 on Linux, and marks nothing on Windows while logging an error for every socket. The commit message says the option is the whole byte "because the six DSCP bits and the two ECN bits share it", but the ECN bits are the part the kernel owns, and the three behaviours above are what happens when an application sets them.

I'd suggest `setDscp(int dscp)` taking 0..63, with the helper writing `dscp << 2`. That behaves the same on all three, it's the value operators configure, and it maps directly onto the Windows QoS mechanisms (`New-NetQosPolicy -DSCPAction`, qWAVE `QOSSetOutgoingDSCPValue`) if we ever add a Windows backend. The ECN-specific tests can then go away.

## 4. Windows marking: a documentation note

For what it's worth, I checked real traffic from Windows 11 (build 26200) with tcpdump on the far side. `IP_TOS` set before `connect()` marks the SYN and data, a listener marked before `listen()` marks the SYN-ACK and is inherited by accepted sockets, and a change after connect takes effect straight away. So the ordering in this PR is fine on Windows.

Microsoft does document `IP_TOS` as unsupported ("use the QoS API"). Older Windows versions ignored it, and a matching QoS policy overrides it, and in those cases `getsockopt` still reads back the requested value. Could the doc comment on `setTos()` say that on Windows this is best effort, and that policy-based QoS (`New-NetQosPolicy`) is the supported way to guarantee marking? Also worth a sentence: an application's own `AfterSocketCreationFunc` is re-invoked on every socket when the value changes, which matters if that callback uses qWAVE and isn't safe to call twice.

## Minor

- `applySocketOptions()` stores `mAppliedTos` even when `setSocketTos()` returned -1. It only costs an extra undo pass later, but checking the return value would keep it accurate.
- `applyTosToOwnSockets()` is public but documented as safe only before the processing thread starts. Protected (it's reachable from the tests via `using`) would prevent misuse.

mykhailopopivniak:
1. Fixed. Both helpers log the error number instead of strerror.

2. Fixed. char* casts at the five sites, initNetwork() at the top of both test
mains, and both sockets are bound before the option is set, which is what every
transport does -- InternalTransport binds before it calls applySocketOptions().
Three of the four failing checks were the ECN ones and are gone with the byte
API.

3. Taken. setDscp() takes 0..63, and setSocketDscp()/getSocketDscp() hold the
shift in one place. getSocketTos() stays as a read-only byte view, which is what
the log line and a capture comparison need. Two things I found while doing it
that strengthen your case: the accepted range now equals the range that can be
honoured everywhere, and the read-back round-trips, so its return is a usable
success test for the first time.

4. Added, both sentences -- and the Windows CI turned up one more thing worth
saying there. Windows has no IPV6_TCLASS: setsockopt refuses it with
WSAENOPROTOOPT, bound or unbound, so a V6 transport gets no marking at all
rather than best-effort marking. The doc comment says so, and testSocketTos
guards its V6 checks the way it already guards the IP_TOS-on-AF_INET6 decoy. A
V6 transport on Windows therefore reports one error per reload, from the
listener read-back, and nothing per socket.

Minor 1: the member is now a bool -- whether a mark was requested -- so there is
nothing to keep accurate. Whether the class took effect is answered by the
listener read-back, which compares what the socket carries against what was
asked and reports a mismatch.

Minor 2 is not done: applyDscpToOwnSockets() can leave the public surface once
the direct start-up call goes, but that marks the listener a cycle later, so a
SYN/ACK already in the backlog could go out unmarked. Better on its own.

sgodin:
A few additional comments:

## 1. Windows does support IPV6_TCLASS here

`rutil/Socket.hxx:131` and the comment at `rutil/test/testSocketTos.cxx:78` say Windows has no `IPV6_TCLASS` and refuses it with WSAENOPROTOOPT (10042). On Windows 11 build 26200 it works:

- `setsockopt(IPPROTO_IPV6, IPV6_TCLASS, 160)` succeeds on both UDP and TCP sockets and reads back 160.
- A receiver using `IPV6_RECVTCLASS` over `::1` sees traffic class 160 on the datagram.
- Through the stack, `setDscp(40)` on V6 UDP and TCP transports leaves the listener and the accepted connection carrying 160, read back with your own `getSocketTos(fd, V6)`.

Which Windows version returned 10042 for you? If it's version-dependent, the doc should name the versions rather than say Windows has no `IPV6_TCLASS`.

In the test, the `#ifndef WIN32` currently compiles out every V6 check, but only the decoy check is platform-specific: Windows refuses `IP_TOS` on an `AF_INET6` socket, and `getsockopt(IP_TOS)` there returns the traffic class. I'd guard just the decoy and the "left the IPv4 level untouched" check. If some Windows versions really do lack the option, the V6 checks could skip at runtime when `setsockopt` fails with WSAENOPROTOOPT, instead of being compiled out.


## 2. Stale reference to onReload()

The doc comment on `applyDscpToOwnSockets()` (`resip/stack/Transport.hxx:429`) still says to "ask through onReload() instead" once the transport is running. `onReload()` no longer schedules anything; that should say `setDscp()`.

## 4. Report when nothing is marked

`setDscp()` raises `mDscpReportPending` on every call, including `-1` on a transport that never marked anything. A service that calls `setDscp(-1)` on every reload because its option is unset then does a `getsockopt` and logs an Info "signaling DSCP 0" line per transport, per reload, for a feature it isn't using. Returning early from `reportDscp()` when `effectiveDscp() < 0` would keep such a service silent.

One request on process: when you address these, could you push them as new commits on top of the branch rather than rebasing and force-pushing a single rewritten commit? Each force-push replaces the whole change, which makes it hard to see what changed between rounds and whether earlier feedback was addressed. The commits can be squashed at merge time.

mykhailopopivniak:
1. IPV6_TCLASS. On Windows Server 2025 (build 10.0.26100, what windows-latest runs) setsockopt(IPV6_TCLASS) returns WSAENOPROTOOPT on UDP and TCP, bound or unbound. Microsoft doesn't list IPV6_TCLASS as settable, so I didn't pin versions — the comment now says the constant is in the SDK and whether the stack honours it differs between builds. Per your second suggestion the V6 checks run everywhere and skip at runtime on WSAENOPROTOOPT; only the decoy and its IPv4-level check stay compiled out.

2. Stale onReload(). Fixed — the comment says setDscp().

3. Report when nothing is marked. Fixed — reportDscp() returns before the read-back when nothing is configured, so an unused transport is silent on every reload. A reset from a configured value still logs once (reporting 0), so removing a marking stays visible.

Round 2 is a commit on top, as you asked.

sgodin:
@mykhailopopivniak Looks great - thanks!

